PR #24, opened 2019-02-27
Adding New Camera Speed Slider

jawchewa:
Adding a Camera Speed Slider to the viewer settings. I also hooked up
the slider to also update on the mouse wheel event, so that the camera
speed can still be set like it was previously, or by using this new
slider. Because of the addition of this slider, the max camera speed is
1000 normally, or 5000 if you are holding shift. Max speed could be adjusted if we want to make it higher or lower, but I felt like this was a pretty good balance.

### DIFF
--- a/src/ui.ts
+++ b/src/ui.ts
@@ -1473,7 +1473,7 @@ class ViewerSettings extends Panel {
         this.camSpeedSlider = this.contents.querySelector('.CamSpeedSlider');
         this.camSpeedSlider.oninput = this.onCamSliderChange.bind(this);
         this.camSpeedSlider.value = '6'
-        addEventListener('wheel', this._onWheel.bind(this), { passive: false });
+        window.addEventListener('wheel', this._onWheel.bind(this), { passive: false });
 
         this.cameraControllerWASD = this.contents.querySelector('.CameraControllerWASD');
         this.cameraControllerWASD.onclick = () => {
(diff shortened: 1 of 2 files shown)


src/ui.ts
```diff
@@ -1490,9 +1501,29 @@ class ViewerSettings extends Panel {
...
+    };
+
     private setCameraControllerClass(cameraControllerClass: CameraControllerClass) {
         this.viewer.setCameraController(new cameraControllerClass());
         this.cameraControllerSelected(cameraControllerClass);
+        this.updateCameraSpeed();
+    }
+
+    private updateCameraSpeed(): void {
+        const controller: FPSCameraController = (this.viewer.cameraController as FPSCameraController);
```
magcius: Could you add setKeyMoveSpeed as an interface to CameraController, and make it an empty function in the Orbit one?

src/ui.ts
```diff
@@ -1464,6 +1470,11 @@ class ViewerSettings extends Panel {
         this.fovSlider.oninput = this.onFovSliderChange.bind(this);
         this.fovSlider.value = '25';
 
+        this.camSpeedSlider = this.contents.querySelector('.CamSpeedSlider');
+        this.camSpeedSlider.oninput = this.onCamSliderChange.bind(this);
+        this.camSpeedSlider.value = '6'
+        addEventListener('wheel', this._onWheel.bind(this), { passive: false });
```
magcius: Please explicitly make this `window.addEventListener` rather than relying on the global.

PR #32, opened 2019-02-28
WW Render Hacks Enhancements

jawchewa:
This PR is to improve the render hacks options for wind waker by doing 2 things:

1. Adds a new Enable Objects toggle that shows/hides all of the models in the scene that are not part of the main map mesh. This would make it so people can see just the main map without all of the actors scattered through it.

2. Applies "Enable Vertex Colors" and "Enable Textures" Render Hacks to all objects in the scene. Previously, these options weren't getting applied to most of the non map models, so characters would still have their textures while everything else in the scene wouldn't. This changes it so those options are applied to pretty much all models instead.


magcius:
Nice work!

PR #38, opened 2019-03-03
Fixing Edge Whitescreen

jawchewa:
Making new util function to get around edge whitescreen.


magcius:
Manually merged this one, since it was tricky.

PR #47, opened 2019-03-11
Add Basic Controller Support

jawchewa:
This branch gives user the ability to navigate maps using a controller. It's primarily setup to work with XBox One and Playstation 4 controllers, so other kinds of controllers may not be fully supported. This could definitely be improved in the future with more controller functionality, and better control, but this is more of a starting point/proof of concept.

If a controller is not immediately recognized, it may need to be disconnected and reconnected before it is picked up.

Current Xbox One Controls:
Left Stick - Move
Right Stick - Rotate Camera
Left Trigger and Right Trigger - Sprint/Speed increase
A and B - Rise and Fall
Start - Pause


src/InputManager.ts
```diff
@@ -203,4 +230,33 @@ export default class InputManager {
...
+
+    private disconnecthandler(e: GamepadEventInit): void {
+        delete this.gamepads[e.gamepad.index];
+    }
+
+    public getGamepadAxis(axis: number): number {
+        let value = 0;
+        this.gamepads.forEach(gamepad => {
+            if(gamepad.connected){
+                if(gamepad.mapping != 'standard' && axis >= 3) axis += 2;
```
jawchewa: This line is kind of a hack, to make it work with PS4 controllers. It may potentially cause problems with other types of non-standard third party controllers, but without it, PS4 control wouldn't work.

magcius:
So, I've held off on merging this, because I felt a "camera rewrite" was coming soon. That did not happen. And given that I'm about to merge #190, I feel like it's probably not going to happen unless I do it.

If you want to rebase this, I will merge it.

src/InputManager.ts
luser: FYI you should be able to check the `Gamepad.id` field for the right vendor and product ID to apply this only for PS4 controllers. Here's some extremely old sample code I wrote back when I was actively working on the Gamepad API:
https://github.com/luser/gamepadmapping.js/blob/6dae0a5751df4acacfae2a5d3619758b7bb4bf43/gamepadmapping.js#L133

PR #49, opened 2019-03-17
ZSI Environment Setting Support

meckzqz:
This fork features basic environment setting interpretation from Grezzo's modified version of the format, rudimentary shader edits to render lighting and fog, as well as some render hacks that I was experimenting with and was too lazy to remove.

Some current issues with this code:

- MM3D's environment setting format seems to differ from OOT3D. This code only seems to work properly with OOT3D as a result.
- Data found at the beginning and end of each environment setting is currently unknown and unused.
- Total number of environment settings seem to report incorrectly in most cases, so for now the first one is the only one read to avoid out of index exceptions.
- Lighting information is only being applied to room renderers.
- Fog start and draw distance don't seem to be accurate when comparing with hardware. Fog algorithm may not be linear?


src/bk/render.ts
```diff
@@ -176,6 +176,8 @@ class DrawCallInstance {
     private texturesEnabled = true;
     private monochromeVertexColorsEnabled = false;
     private alphaVisualizerEnabled = false;
+    private vertexNormalsEnabled = false;
```
magcius: Why did you modify the Banjo-Kazooie files?

src/oot3d/render.ts
```diff
@@ -274,13 +274,37 @@ in vec2 v_TexCoord0;
...
+in float v_FogStart;
+
 void main() {
     ${this.generateTextureEnvironment(this.material.textureEnvironment)}
 
     if (!(${this.generateAlphaTestCompare(this.material.alphaTestFunction, this.material.alphaTestReference)}))
         discard;
 
-    gl_FragColor = t_CmbOut;
+    vec4 ResultColor = t_CmbOut;
```
magcius: Temporaries in shader code should start with `t_`

src/oot3d/render.ts
```diff
@@ -274,13 +274,37 @@ in vec2 v_TexCoord0;
...
+
+    #ifdef USE_VERTEX_NORMAL
+        ResultColor.rgb = normalize(v_Normal) * 0.5 + 0.5; 
+    #endif
+
+    #ifdef USE_UV
+        ResultColor.r = v_TexCoord0.x;
+        ResultColor.g = v_TexCoord0.y;
+        ResultColor.b = 0.0;
```
magcius: I personally prefer my USE_UV to have b = 1.0.

src/oot3d/render.ts
```diff
@@ -274,13 +274,37 @@ in vec2 v_TexCoord0;
...
+    #endif
+
+    #ifdef USE_LIGHTING
+        float FogFactor = clamp((v_DrawDistance - v_Depth) / (v_DrawDistance - v_FogStart), 0.0, 1.0);
+        ResultColor.rgb = mix(v_FogColor, t_CmbOut.rgb * v_Lighting, FogFactor);
```
magcius: Put this before the other two ifdef's so it's more clear it's a hack. This is also a bit borked -- I'll fix it at some point.

src/oot3d/render.ts
```diff
@@ -427,15 +481,90 @@ class MaterialInstance {
...
+        this.uvEnabled = v;
+        this.createProgram();
+    }
+
+    public setLightingEnabled(v: boolean): void {
+        this.lightingEnabled = v;
+        this.createProgram();
+    }
+
+    public setAmbientColor(color: vec3): void {
```
magcius: Perhaps have this all as one giant function call?

```js
public setEnvironmentSettings(ambientColor: vec3, primaryLightColor: vec3, ...)
```

src/oot3d/render.ts
```diff
@@ -407,6 +448,19 @@ class MaterialInstance {
...
+    private uvEnabled: boolean = false;
+
+    public ambientLightCol: vec3;
+    public primaryLightCol: vec3;
+    public primaryLightDir: vec3;
+    public secondaryLightCol: vec3;
+    public secondaryLightDir: vec3;
+    public fogCol: vec3;
+    public fogStart: number;
+    public drawDistance: number;
```
magcius: Could you give these all defaults rather than relying on `undefined` and patching in the generated shader code?

```js
    public ambientLightColor = vec3.create();
    public primaryLightColor = vec3.create();
    public fogStart = 0.0;
```

etc.

src/oot3d/render.ts
```diff
@@ -357,6 +389,15 @@ void main() {
 
     v_Color = a_Color;
 
+    v_Normal = a_Normal;
+    v_Depth = gl_Position.w;
```
magcius: OK, you're using clip-space w. Try using eye-space w instead -- it might give you better results. Of course, it's incredibly inaccurate to how the 3DS works. I'll clean this up at some point.

src/oot3d/render.ts
meckzqz: defaults are now defined in the class itself, though we could still do this in render.ts with a null check if it is preferable. i commented out the later solution and pushed both.

src/oot3d/render.ts
```diff
@@ -544,14 +502,30 @@ class MaterialInstance {
...
+        //    tempEnvironmentSettings.drawDistance = 0.0;
+        //}
+
+        additionalParameters += `vec3 AMBIENT_LIGHT_COLOR = vec3(${tempEnvironmentSettings.ambientLightCol});\n`;
+        additionalParameters += `vec3 PRIMARY_LIGHT_COLOR = vec3(${tempEnvironmentSettings.primaryLightCol});\n`;
+        additionalParameters += `vec3 PRIMARY_LIGHT_DIRECTION = vec3(${tempEnvironmentSettings.primaryLightDir});\n`;
+        additionalParameters += `vec3 SECONDARY_LIGHT_COLOR = vec3(${tempEnvironmentSettings.secondaryLightCol});\n`;
+        additionalParameters += `vec3 SECONDARY_LIGHT_DIRECTION = vec3(${tempEnvironmentSettings.secondaryLightDir});\n`;
+        additionalParameters += `vec3 FOG_COLOR = vec3(${tempEnvironmentSettings.fogCol});\n`;
+        additionalParameters += `float FOG_START = ${tempEnvironmentSettings.fogStart + ".0"};\n`;
```
magcius: if fogStart has a non-zero fractional component this will break. Use `this.generateFloat`. make it protected / public if you need to.

src/oot3d/render.ts
```diff
@@ -544,14 +502,30 @@ class MaterialInstance {
...
-        additionalParameters += `vec3 PRIMARY_LIGHT_DIRECTION = vec3(${this.primaryLightDir ? this.primaryLightDir : "0, 0, 0"});\n`;
-        additionalParameters += `vec3 SECONDARY_LIGHT_COLOR = vec3(${this.secondaryLightCol ? this.secondaryLightCol : "0, 0, 0"});\n`;
-        additionalParameters += `vec3 SECONDARY_LIGHT_DIRECTION = vec3(${this.secondaryLightDir ? this.secondaryLightDir : "0, 0, 0"});\n`;
-        additionalParameters += `vec3 FOG_COLOR = vec3(${this.fogCol ? this.fogCol : "0, 0, 0"});\n`;
-        additionalParameters += `float FOG_START = ${this.fogStart ? this.fogStart + ".0" : "0.0"};\n`;
-        additionalParameters += `float DRAW_DISTANCE = ${this.drawDistance ? this.drawDistance + ".0" : "0.0"};\n`;
+        let tempEnvironmentSettings = new ZSI.ZSIEnvironmentSettings();
+
+        if (this.environmentSettings) tempEnvironmentSettings = this.environmentSettings;
+        // NOTE(quade): would it be preferable to define defaults here instead of in the class itself?
```
magcius: nah, what you have is good

src/oot3d/render.ts
```diff
@@ -939,7 +1067,7 @@ export abstract class BasicRendererHelper {
         this.viewRenderer.prepareToRender(device);
 
         this.renderTarget.setParameters(device, viewerInput.viewportWidth, viewerInput.viewportHeight);
-        const finalPassRenderer = this.renderTarget.createRenderPass(device, standardFullClearRenderPassDescriptor);
+        const finalPassRenderer = this.renderTarget.createRenderPass(device, makeClearRenderPassDescriptor(true, colorNew(0, 0, 0, 1)));
```
magcius: Hm, why was this changed?

src/oot3d/zsi.ts
```diff
@@ -67,13 +80,91 @@ function readActors(version: Version, buffer: ArrayBufferSlice, nActors: number,
...
     const roomSize = version === Version.Ocarina ? 0x44 : 0x34;
+
     for (let i = 0; i < nRooms; i++) {
         rooms.push(readString(buffer, offs, roomSize));
         offs += roomSize;
     }
     return rooms;
 }
 
+function decimalToHexString(number: number)
```
magcius: Unused; please remove.

PR #78, opened 2019-05-22
 fix black holes and other billboards

jupahe64:


magcius:
Fixed through 4e99eeae6c70c05f10b7207712caee6539a590ec

PR #80, opened 2019-05-29
worldmaps for smg2

jupahe64:


magcius:
This needs to use the new LiveActor system; building on the legacy renderer system with ObjectBase is gonna land you in hot water mostly.

jupahe64:
Alright I guess I will rewrite the whole code then, only keeping the loading and some rendering code

PR #114, opened 2019-09-15
Pilotwings 64: Parse and render model files.

PFedak:
Doesn't work right now, but the model parsing is correct at least.


PFedak:
Ok now they render fine, but I haven't figured out where all the scaling info lives yet.

PR #117, opened 2019-09-17
Pilotwings 64: Texture things mostly correctly

PFedak:
The actually texture mapping is nowhere near correct yet.

I changed MeshInstance into MeshRenderer based on the bk naming, and it now has a collection of MaterialInstance s (not married to the name) which actually do the rendering.

I added a method to the TextureHolder that probably shouldn't be there, but it seemed preferable to unprotecting the equivalent method or adding an index->address lookup table, since everything uses those indices instead of addresses. 

I could also leave out the dummy texture, I just thought it would be visually helpful.


src/Pilotwings64/Scenes.ts
```diff
@@ -737,7 +745,7 @@ vec4 Texture2D_N64_Bilerp(sampler2D t_Texture, vec2 t_TexCoord) {
     return t_S0 + abs(t_Offs.x)*(t_S1-t_S0) + abs(t_Offs.y)*(t_S2-t_S0);
 }
 
-#define Texture2D_N64 Texture2D_N64_Point
+#define Texture2D_N64 Texture2D_N64_Bilerp
```
magcius: This should eventually come from G_OTHERMODE_H. See gDPSetTextureFilter and friends.

PR #118, opened 2019-09-18
Pilotwings 64: Add crazy combine shader.

PFedak:
Also change uses of "blend" to "combine" to be more accurate.

I had to do some weird things to smuggle an integer array into the shader: I got an error about precision mismatch between the vertex and fragment shaders if i used a uvec directly and got weird data integrity errors if i moved the uniform buffer definition to just the fragment shader. I'm hoping you can show me the real way.

By "data integrity" errors, I mean the shader output would change dramatically if I resized the window, not really sure how that's possible.

### DIFF
--- a/src/Pilotwings64/Scenes.ts
+++ b/src/Pilotwings64/Scenes.ts
@@ -13,8 +13,8 @@ import { decompress } from "../compression/MIO0";
 import { makeStaticDataBuffer } from "../gfx/helpers/BufferHelpers";
 import { DeviceProgram } from "../Program";
 import { GfxRenderInstManager } from "../gfx/render/GfxRenderer";
-import { fillMatrix4x3, fillMatrix4x4, fillMatrix4x2 } from "../gfx/helpers/UniformBufferHelpers";
-import { mat4, vec3 } from "gl-matrix";
+import { fillMatrix4x3, fillMatrix4x4, fillMatrix4x2, fillVec4v, fillVec4 } from "../gfx/helpers/UniformBufferHelpers";
+import { mat4, vec3, vec4 } from "gl-matrix";
 import { GfxRenderHelper } from "../gfx/render/GfxRenderGraph";
 import { standardFullClearRenderPassDescriptor, BasicRenderTarget } from "../gfx/helpers/RenderTargetHelpers";
 import { computeViewMatrix } from "../Camera";
@@ -329,6 +329,13 @@ interface UV_Scroll {
     scaleT: number;
 }
 
+interface CombineParams {
+    a: number;
+    b: number;
+    c: number;
+    d: number;
+}
+
 interface UVTX {
     name: string;
     width: number;
@@ -338,9 +345,13 @@ interface UVTX {
     levels: UVTX_Level[];
     cms: number;
     cmt: number;
-    blendLayer: number;
+    combine: CombineParams[];
+
+    combineIndex?: number;
     uvScroll?: UV_Scroll;
-    blendScroll?: UV_Scroll;
+    combineScroll?: UV_Scroll;
+    primitive?: vec4;
+    environment?: vec4;
 }
 
 function parseUVTX_Chunk(chunk: Pilotwings64FSFileChunk, name: string): UVTX {
@@ -350,10 +361,12 @@ function parseUVTX_Chunk(chunk: Pilotwings64FSFileChunk, name: string): UVTX {
 
     const scaleS = view.getFloat32(0x04);
     const scaleT = view.getFloat32(0x08);
-    const blendScaleS = view.getFloat32(0x0c);
-    const blendScaleT = view.getFloat32(0x10);
-    const uvScroll: UV_Scroll = { scaleS, scaleT };
-    const blendScroll: UV_Scroll = { scaleS: blendScaleS, scaleT: blendScaleT };
+    const combineScaleS = view.getFloat32(0x0c);
+    const combineScaleT = view.getFloat32(0x10);
+
+    let primitive: vec4 | undefined;
+    let environment: vec4 | undefined;
+    const combine: CombineParams[] = [];
 
     const textureState = new TextureState();
     const tiles: TileState[] = nArray(8, () => new TileState());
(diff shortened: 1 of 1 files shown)


src/Pilotwings64/Scenes.ts
```diff
@@ -377,6 +390,31 @@ function parseUVTX_Chunk(chunk: Pilotwings64FSFileChunk, name: string): UVTX {
             assert(on);
             textureState.set(on, tile, level, s, t);
         } else if (cmd === F3D_GBI.G_SETCOMBINE) {
+            // many of these 0x07 should be 0x0f, but noise isn't implemented
+            // so this sends noise to 0, along with 0xf (which should be 0)
+            const a0  = (w0 >> 20) & 0x07;
```
magcius: You want triple >>> here for unsigned right-shift.

src/Pilotwings64/Scenes.ts
```diff
@@ -852,24 +976,28 @@ class MeshRenderer {
...
+    constructor(private materialData: MaterialData, textureData: TextureData[]) {
         this.hasTexture = materialData.textureIndex < 0x0FFF;
         if (this.hasTexture) {
             const mainTextureData = textureData[materialData.textureIndex];
             this.uvtx = mainTextureData.uvtx;
             mainTextureData.fillTextureMapping(this.textureMappings[0]);
-            if (this.uvtx.blendLayer < 0x0FFF) {
-                this.hasBlend = true;
-                textureData[this.uvtx.blendLayer].fillTextureMapping(this.textureMappings[1]);
+            if (this.uvtx.combineIndex) {
```
magcius: You probably want `this.uvtx.combineIndex !== undefined` so that you don't accidentally prevent 0 from being a combine index.

src/Pilotwings64/Scenes.ts
```diff
@@ -896,19 +1024,31 @@ class MaterialInstance {
...
-                if (this.uvtx.blendScroll) {
-                    texMatrixScratch[12] = -((viewerInput.time / 1000) * this.uvtx.blendScroll.scaleS) % 1;
-                    texMatrixScratch[13] = -((viewerInput.time / 1000) * this.uvtx.blendScroll.scaleT) % 1;
+                if (this.uvtx.combineScroll) {
+                    texMatrixScratch[12] = -((viewerInput.time / 1000) * this.uvtx.combineScroll.scaleS) % 1;
+                    texMatrixScratch[13] = -((viewerInput.time / 1000) * this.uvtx.combineScroll.scaleT) % 1;
                 }
                 offs += fillMatrix4x2(d, offs, texMatrixScratch);
-                this.program.defines.set('HAS_BLEND', '1');
+                this.program.defines.set('HAS_COMBINE', '1');
```
magcius: This is... sort of misleading. Materials with one texture can still have a combine -- they're just a 1-cycle combine. If you have two textures, it's likely it's a 2-cycle combine.

src/Pilotwings64/Scenes.ts
```diff
@@ -771,25 +854,66 @@ vec4 Texture2D_N64_Bilerp(sampler2D t_Texture, vec2 t_TexCoord) {
...
 }
 
+vec3 combineColorCycle(vec4 combColor, vec4 tex0, vec4 tex1, float params) {
+
+    vec3 colorInputs[8] = vec3[8](
+        combColor.rgb, tex0.rgb, tex1.rgb, u_PrimColor.rgb,
+        v_Color.rgb, u_EnvColor.rgb, one.rgb, zero.rgb
+    );
+
+    vec3 multInputs[16] = vec3[16](
```
magcius: multInputs is unused? That doesn't seem right. Quite sure you want `*multInputs[p.z]`

src/Pilotwings64/Scenes.ts
```diff
@@ -771,25 +854,66 @@ vec4 Texture2D_N64_Bilerp(sampler2D t_Texture, vec2 t_TexCoord) {
...
+        combColor.rgb, tex0.rgb, tex1.rgb, u_PrimColor.rgb,
+        v_Color.rgb, u_EnvColor.rgb, zero.rgb /* key */, combColor.aaa,
+        tex0.aaa, tex1.aaa, u_PrimColor.aaa, v_Color.aaa,
+        u_EnvColor.aaa, zero.rgb /* LOD */, zero.rgb /* prim LOD */, zero.rgb
```
magcius: LOD / prim LOD are used for mipmapping, btw. the two tiles chosen are the two tiles of the mipmap. It's probably set up specially by the engine.

src/Pilotwings64/Scenes.ts
```diff
@@ -852,24 +976,28 @@ class MeshRenderer {
     }
 }
 
+function floatify(params: CombineParams): number {
+    return params.a / 16 + params.b / 256 + params.c / 4096 + params.d / 65536;
```
magcius: You could do `(params.r << 24) | (params.g << 16) | (params.b << 8) | (params.a)`, and then the equivalent bitshifts in the shader after you cast the float to int. That might be less finnicky.

src/Pilotwings64/Scenes.ts
PFedak: A lot of the mip-mapped ones without "combine" are also two cycle (LOD calculation then multiply the result with vertex color), so yeah, I agree, not a good name.

I'll try to come up with a better name, unless you'd prefer it stays "blend".

src/Pilotwings64/Scenes.ts
PFedak: Wow, that's embarrassing. I'm hoping it was used at some point and I just messed it up when I was trying to get things working. It seems like the only higher values used are for 0 (which it was getting anyway, maybe, LOD calculations (which, again, would do nothing), and getting the "shaded" alpha value, which only seems to be used at the edge of water.

src/Pilotwings64/Scenes.ts
PFedak: Ah, I was ruling out sending an integer value because I would lose precision, but if it's packed better that's not an issue. I was hoping there was an actual way to send integer data over, though, unless that's just super unusual.

src/Pilotwings64/Scenes.ts
PFedak: Ah good point, thanks

src/Pilotwings64/Scenes.ts
PFedak: I meant to look up what that did

src/Pilotwings64/Scenes.ts
magcius: numbers in JavaScript are doubles, which means you can represent 53-bit integer values without any loss of precision.

src/Pilotwings64/Scenes.ts
magcius: USE_2CYCLE maybe?

src/Pilotwings64/Scenes.ts
PFedak: Maybe I'm not understanding what you mean by "cast the float to int" then. Is there a way to get around it being a float32 in the shader? I only know read up on enough GLSL to get this working.

src/Pilotwings64/Scenes.ts
PFedak: I went with calling it a "paired texture", since it seems like the documentation calls the LOD stuff "two-cycle" as well.

src/Pilotwings64/Scenes.ts
magcius: er, right, since we stuff these in 32-bit floats, we only have 23 bits of mantissa. Thankfully, A/B/D are 3-bit, and C is 4-bit, so we could just do: `return (params.a << 10) | (params.b << 7) << (params.c << 3) | (params.d);`

magcius: In GLSL, you can cast back to an int with `int(params.x)` or similar, and then do similar bit-shifting stuff.

src/Pilotwings64/Scenes.ts
PFedak: (Also, as long as we're sending the same texture data for 0 and 1 on the mipmapped ones leaving these zero won't affect anything)

PR #120, opened 2019-09-19
Pilotwings 64: Fix lots of small texture issues.

PFedak:
- Set scrolling texture dimensions correctly
- Fix I4 and IA4 decoding
- Properly handle paired textures in tile 0
- Check 2-cycle and filtering "other" modes

### DIFF
--- a/src/Common/N64/Image.ts
+++ b/src/Common/N64/Image.ts
@@ -137,7 +137,7 @@ export function decodeTex_CI8(dst: Uint8Array, view: DataView, srcIdx: number, t
 export function decodeTex_IA4(dst: Uint8Array, view: DataView, srcOffs: number, tileW: number, tileH: number, line: number = 0, deinterleave: boolean = false): void {
     let dstIdx = 0;
     let srcIdx = 0;
-    const padW = (line !== 0) ? (((line << 4) - tileW) << 1) : 0x00;
+    const padW = (line !== 0) ? (((line << 4) - tileW) >>> 1) : 0x00;
     for (let y = 0; y < tileH; y++) {
         const di = deinterleave ? ((y & 1) << 2) : 0;
         for (let x = 0; x < tileW; x += 2) {
@@ -164,7 +164,7 @@ export function decodeTex_IA4(dst: Uint8Array, view: DataView, srcOffs: number, 
 export function decodeTex_IA8(dst: Uint8Array, view: DataView, srcOffs: number, tileW: number, tileH: number, line: number = 0, deinterleave: boolean = false): void {
     let dstIdx = 0;
     let srcIdx = 0;
-    const padW = (line !== 0) ? (((line << 3) - tileW) << 1) : 0x00;
+    const padW = (line !== 0) ? ((line << 3) - tileW) : 0x00;
     for (let y = 0; y < tileH; y++) {
         const di = deinterleave ? ((y & 1) << 2) : 0;
         for (let x = 0; x < tileW; x++) {
@@ -205,7 +205,7 @@ export function decodeTex_IA16(dst: Uint8Array, view: DataView, srcOffs: number,
 export function decodeTex_I4(dst: Uint8Array, view: DataView, srcOffs: number, tileW: number, tileH: number, line: number = 0, deinterleave: boolean = false): void {
     let dstIdx = 0;
     let srcIdx = 0;
-    const padW = (line !== 0) ? (((line << 4) - tileW) << 1) : 0x00;
+    const padW = (line !== 0) ? (((line << 4) - tileW) >>> 1) : 0x00;
     for (let y = 0; y < tileH; y++) {
         const di = deinterleave ? ((y & 1) << 2) : 0;
         for (let x = 0; x < tileW; x += 2) {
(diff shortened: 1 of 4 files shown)


PFedak:
Notably waterfalls and the american southwest now look good.

I changed some things in shared files - the slight otherModeH refactoring should be safe, but the image decoding changes have the potential to break something I guess, though they felt like typos to me.

src/Common/N64/Image.ts
```diff
@@ -214,12 +214,12 @@ export function decodeTex_I4(dst: Uint8Array, view: DataView, srcOffs: number, t
             dst[dstIdx + 0] = i0;
             dst[dstIdx + 1] = i0;
             dst[dstIdx + 2] = i0;
-            dst[dstIdx + 3] = i0;
+            dst[dstIdx + 3] = 0xff;
```
magcius: Quite sure this is wrong -- intensity should be all four channels.

magcius: See https://github.com/gonetz/GLideN64/blob/36078a05014426822990fe202ecf7a190fca5dae/src/convert.h#L341-L362

PFedak: Yeah this was the line I'm least sure about. It definitely makes things look better here but that might just because there's supposed to be a dark skybox (water?) behind it, which would probably have the same effect.

src/Pilotwings64/Scenes.ts
```diff
@@ -1267,6 +1331,7 @@ class Pilotwings64SceneDesc implements SceneDesc {
                 return parseUVTX(file);
             } catch (e) {
                 // preserve the ordering of the textures for indexing
+                console.warn(file.name, e)
```
magcius: Are any textures failing to load still?

src/bk/f3dex.ts
```diff
@@ -164,13 +164,43 @@ export class RSPOutput {
...
+    return (modeH >>> OtherModeH_Layout.G_MDSFT_CYCLETYPE) & 0x03;
+}
+
+export const enum OtherModeH_TextFilt {
+    G_TF_POINT   = 0x00,
+    G_TF_AVERAGE = 0x03,
+    G_TF_BILERP  = 0x02,
+}
+
+export function translateTextFilt(modeH: number): OtherModeH_TextFilt {
```
magcius: Typically "translate" means "convert to a Gfx thing", so I would prefer to see this called something like "getTextFiltFromOtherModeH".

src/Common/N64/Image.ts
PFedak: http://n64devkit.square7.ch/pro-man/pro13/13-08.htm#02 documentation supports you, too. I'll just have to wait until the skybox is in.

src/bk/f3dex.ts
```diff
@@ -164,13 +164,43 @@ export class RSPOutput {
...
     G_CYC_COPY   = 0x02,
     G_CYC_FILL   = 0x03,
 }
 
+export function translateCycleType(modeH: number): OtherModeH_CycleType {
+    return (modeH >>> OtherModeH_Layout.G_MDSFT_CYCLETYPE) & 0x03;
+}
+
+export const enum OtherModeH_TextFilt {
+    G_TF_POINT   = 0x00,
```
magcius: This enum is already defined in `../N64/Image.ts` -- mind changing it to use this?

src/Pilotwings64/Scenes.ts
PFedak: No, not after the I4 change. I can take this out if you'd prefer.

src/bk/f3dex.ts
```diff
@@ -164,13 +164,43 @@ export class RSPOutput {
...
+
 export const enum OtherModeH_CycleType {
     G_CYC_1CYCLE = 0x00,
     G_CYC_2CYCLE = 0x01,
     G_CYC_COPY   = 0x02,
     G_CYC_FILL   = 0x03,
 }
 
+export function translateCycleType(modeH: number): OtherModeH_CycleType {
+    return (modeH >>> OtherModeH_Layout.G_MDSFT_CYCLETYPE) & 0x03;
```
magcius: I think OTHERMODE_H is hardcoded in the RDP, but can you triple-check that it doesn't change between F3D/F3DEX/F3DEX2?

src/bk/f3dex.ts
PFedak: I didn't think to look, whoops

src/bk/f3dex.ts
PFedak: Sounds good to me

src/bk/f3dex.ts
PFedak: gliden64 treats them as compatible (at least, as far as I can tell from reading https://github.com/gonetz/GLideN64/search?q=gSPSetOtherMode_H&unscoped_q=gSPSetOtherMode_H)

src/Pilotwings64/Scenes.ts
magcius: If no textures are failing, I'd prefer to just remove the dummy fallback entirely.

PR #124, opened 2019-09-21
Pilotwings 64: Implement basic blending modes.

PFedak:
The current data doesn't use very many modes, so the mapping
to Gfx modes is mostly straightforward so far.

The render passes are just an approximation for now.

It's probably time to start splitting this file up. Some of the enums could be moved to the f3dex file, though the geo flags definitely changed. There are definitely still some mismatches between my intuition and how things are actually showing up (decals still seem to be z-fighting, etc.).

### DIFF
--- a/src/Pilotwings64/Scenes.ts
+++ b/src/Pilotwings64/Scenes.ts
@@ -1,9 +1,8 @@
 
 import {
-    GfxDevice, GfxBuffer, GfxInputLayout, GfxInputState, GfxBufferUsage,
-    GfxVertexAttributeDescriptor, GfxFormat, GfxVertexAttributeFrequency, GfxRenderPass,
-    GfxHostAccessPass, GfxBindingLayoutDescriptor, GfxTextureDimension, GfxWrapMode,
-    GfxMipFilterMode, GfxTexFilterMode, GfxSampler, GfxBlendFactor, GfxBlendMode, GfxTexture
+    GfxDevice, GfxBuffer, GfxInputLayout, GfxInputState, GfxBufferUsage, GfxVertexAttributeDescriptor, GfxFormat, GfxVertexAttributeFrequency,
+    GfxRenderPass, GfxHostAccessPass, GfxBindingLayoutDescriptor, GfxTextureDimension, GfxWrapMode, GfxMipFilterMode, GfxTexFilterMode,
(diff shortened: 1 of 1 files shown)


PFedak:
It seems `terrainLike` was a bad name - the reality is almost completely the opposite. Not sure how to reconcile that with the effect on render mode.

src/Pilotwings64/Scenes.ts
```diff
@@ -1024,6 +1061,286 @@ function calcScaleForShift(shift: number): number {
...
+
+const enum BlendParam_B {
+    G_BL_1MA   = 0,
+    G_BL_A_MEM = 1,
+    G_BL_1     = 2,
+    G_BL_0     = 3,
+}
+
+interface decodeMaterialResult{
+    geoMode: number,
```
magcius: These should be semicolons, not commas.

PFedak: I guess I got caught up with all the enums. I'm a little surprised the formatter/linter didn't fix that, but I guess I haven't looked at the settings.

src/Pilotwings64/Scenes.ts
```diff
@@ -1389,6 +1390,10 @@ class MaterialInstance {
         computeViewMatrix(scratchMatrix, viewerInput.camera);
         mat4.mul(scratchMatrix, scratchMatrix, modelMatrix);
 
+        if (zmode === ZMode.ZMODE_XLU) {
+            renderInst.sortKey = scratchMatrix[14];
```
magcius: You probably want something like: ```ts
if (zmode === ZMode.ZMODE_XLU) {
    renderInst.sortKey = makeSortKey(GfxRendererLayer.TRANSLUCENT);
    renderInst.sortKey = setSortKeyDepth(renderInst.sortKey, scratchMatrix[14]);
} else {
    renderInst.sortKey = makeSortKey(GfxRendererLayer.OPAQUE);
}
```

PFedak: Oh, would I do that instead of multiple passes? Seems easy enough. I'm assuming I should still keep decals separate from other surfaces, unless we're just embracing the z-fighting for now.

PFedak:
Ok, finally merging because I fixed two errors (misread the direction of an if statement, and missed the importance of CVG_X_ALPHA on the coverage value itself).

PR #134, opened 2019-09-26
Pilotwings64: Start on Dobj support: boats!

PFedak:
Adding more will be rather painstaking, hopefully this framework
is robust enough for the other movement types.

I'm doing some maybe-iffy things with optional/undefined here just because I don't know the language well enough to be aware of alternatives.


PFedak:
It turns out the hang gliders use the same motion.

I made one small change to the way ObjectRenderer works to let the dynamic objects better reuse that framework. The alternative I considered was just passing in the toNoclip transform as a parent.

PR #140, opened 2019-10-01
Add GTA3

davidar:
This adds basic support for rendering GTA3 map areas. There's still some rendering issues (mostly related to ~~transparency and~~ lighting), and it needs some performance optimisation. Parsing is done with a [WebAssembly build of librw](https://github.com/davidar/librw-node), but it's only loaded when needed so shouldn't add any overhead to the loading time of noclip in general.

image


davidar:
:tada:

PR #141, opened 2019-10-02
Pilotwings64: Add task objects.

PFedak:
Based mostly on roto's work. Rings still have some weird things
going on (and seem to all have the clock hands). Scaling on the
thermals was just a guess.

The messy landing pad logic could might better be done by just allowing
the two landing pads to draw at the same time, knowing the active one
will be drawn second.


src/Pilotwings64/Scenes.ts
```diff
@@ -2657,6 +2899,39 @@ class Pilotwings64Renderer implements SceneGfx {
         return skyPassRenderer;
     }
 
+    public setCurrentTask(index: number): void {
```
PFedak: sorry this got so big, I'm mainly concerned if I messed up something non-obvious in this section.

src/Pilotwings64/Scenes.ts
```diff
@@ -2657,6 +2899,39 @@ class Pilotwings64Renderer implements SceneGfx {
...
+            const taskNumber = this.strIndexToTask[strIndex - 1];
+            this.setCurrentTask(taskNumber);
+        };
+        this.taskSelect.selectItem(0);
+
+        taskPanel.contents.appendChild(this.taskSelect.elem);
+
+        return [taskPanel];
+    }
+
```
PFedak: sorry this got so big, I wanted to make sure I wasn't missing some subtle thing with the UI

PR #142, opened 2019-10-03
Pilotwings64: Allow weather/time selection.

PFedak:
- Implement texture palette, refactoring how DataHolder is used
 - Load skybox and ocean plane from UVEN
 - Also add back ring angles and fix landing pad bug, whoops


PFedak:
Making a PR since I already have it working this way, might think about other approaches for the UI but I think I'd agree that discoverability is more important than an easy interface for "power users".

PR #143, opened 2019-10-06
GTA3 lighting and improved performance

davidar:
This renders ambient lighting according to simulated time of day and selected weather scenario.


davidar:
Updated with the rewritten renderer, which performs well enough to show the whole map:

image

PR #144, opened 2019-10-07
Pilotwings64: Let it snow!

PFedak:
Wasn't sure if adding another primitive type was a big deal. I looked at doing the particle calculations in the shader but it seems like transform feedback is what I want and that isn't supported on all browsers.


magcius:
As discussed on Discord, I'm not too keen on using GL points, as they have different behavior on different GPUs ( one of the more famous AMD/NVIDIA differences; see https://github.com/KhronosGroup/OpenGL-API/issues/51 ). Point sprites / billboards are a perfectly fine replacement if you want to return to this eventually.

PR #147, opened 2019-10-13
bk: Add vertex effects.

PFedak:


PFedak:
This is still in progress (a handful more animation types, some look pretty involved). I mostly wanted to make sure this was what you had in mind.

I'll make sure to fill out the vertex buffer completely to get rid of the empty slots I have now for unused vertices.

PFedak:
With that last little fix I think this is good to go.

PR #149, opened 2019-10-15
Pilotwings64: Let it snow!

PFedak:
Each snowflake is a quad, smuggling in a corner identifier
for proper billboarding.


PFedak:
I understand the reason to use instances now: you can't compute the billboard transform based only on the coordinates of a corner. The conversion from corner number to offset is a little gross, but it does allow all the computation to be done in the shader.

Playing around with the snow in game made me realize I had a subtle error in the flake motion. Still kind of bizarre but I'm pretty sure this is correct..

PFedak:
Thanks for the tip about gl_VertexID, much simpler now.

PR #150, opened 2019-10-18
bk: Parse file mappings from game data.

PFedak:
Also load the setup files. Three levels differ from the hardcoded values:
- Nabnut's Attic - Fall is identical to Attic 2 instead of Attic - Winter
- Grunty's Lair Floor 5 has a skylight effect over the urn
- Grunty's Lair Floor 8 has another skylight-ish effect


PFedak:
Not sure what the deal with the Nabnut's Attic scenes are, the scene IDs are definitely correct (so possibly different object placements), but either way there are only two unique sets of geometry files among the three scenes.

PR #151, opened 2019-10-19
GrandTheftAutoViceCity

davidar:
The only major technical change here is extending the GTA3 renderer to handle compressed textures, and some improvements for water transparency. There's still some issues with transparent objects being drawn in the wrong order, which I haven't been able to sort out yet.

image


magcius:
Merged manually (had to fix some conflicts)

PR #153, opened 2019-10-26
GrandTheftAutoSanAndreas

davidar:
I think I might need to handle the LODs a bit more intelligently, as there's some glitches related to that (some z-fighting and funky mipmaps when zoomed out). It could probably do with some more optimisation too, as the size of the full map is pushing the draw call count.

image


src/GrandTheftAuto3/item.ts
```diff
@@ -114,15 +125,44 @@ export interface ItemPlacement {
 }
 
 export function parseItemPlacement(text: string): ItemPlacement {
-    let instances = [] as ItemInstance[];
+    const instances = [] as ItemInstance[];
```
magcius: fyi I usually write `const instances: ItemInstance[] = [];`

src/GrandTheftAuto3/program.glsl
```diff
@@ -11,6 +11,7 @@ layout(row_major, std140) uniform ub_SceneParams {
     vec4 u_SkyBotColor;
     vec4 u_WaterColor;
     vec4 u_WaterOrigin;
+    float u_Time;
```
magcius: I'll fix this for you later, but you really shouldn't put floats in uniform blocks, as all fields require 16 byte alignment in std140 layout. I typically use a `vec4 u_Misc` field and then do `#define u_Time (u_Misc.x)`.

src/GrandTheftAuto3/render.ts
```diff
@@ -387,10 +399,10 @@ export class DrawKey {
     constructor(obj: ObjectDefinition, public zone: string) {
         if (obj.flags & ObjectFlags.DRAW_LAST) {
             this.renderLayer = GfxRendererLayer.TRANSLUCENT;
-            this.modelName = obj.modelName;
+            //this.modelName = obj.modelName;
```
magcius: Reason for this?

magcius:
Merged manually.

src/GrandTheftAuto3/render.ts
davidar: I (temporarily) commented that out to reduce the draw call count for the full map, at the expense of possibly causing more transparency sorting issues. Not yet sure of the best way to solve it.

PR #165, opened 2019-11-06
GTA: compress assets

davidar:
This roughly halves the download size for each map. You'll need to run the `compress` script and upload the `.imgz` files it generates.


magcius:
Merged manually.

PR #166, opened 2019-11-06
BanjoKazooie: Start adding objects from setup files.

PFedak:
Just add objects that are always spawned on level load for now,
as others may have a different data format. Even some of these
use a new type of file which I haven't looked into yet.

Add a persistent store for the model data, maybe more later.

Interpret model file associations (and animation metadata) from
the code. The alternative was parsing a blob with lots of different
data, which seemed error prone.

Also fix two small bugs: a render template instance memory leak
and a missing "/2" in the f3dex G_VTX code.

### DIFF
--- a/src/BanjoKazooie/scenes.ts
+++ b/src/BanjoKazooie/scenes.ts
@@ -7,12 +7,17 @@ import * as BYML from '../byml';
 import { GfxDevice, GfxHostAccessPass, GfxRenderPass } from '../gfx/platform/GfxPlatform';
 import { FakeTextureHolder, TextureHolder } from '../TextureHolder';
 import { textureToCanvas, N64Renderer, N64Data, BKPass } from './render';
-import { mat4 } from 'gl-matrix';
+import { mat4, vec3 } from 'gl-matrix';
 import { transparentBlackFullClearRenderPassDescriptor, depthClearRenderPassDescriptor, BasicRenderTarget } from '../gfx/helpers/RenderTargetHelpers';
 import { SceneContext } from '../SceneBase';
 import { GfxRenderHelper } from '../gfx/render/GfxRenderGraph';
 import { executeOnPass } from '../gfx/render/GfxRenderer';
 import { GfxRenderCache } from '../gfx/render/GfxRenderCache';
+import ArrayBufferSlice from '../ArrayBufferSlice';
+import { assert } from '../util';
+import { DataFetcher } from '../DataFetcher';
+import { Endianness } from '../endian';
+import { MathConstants } from '../MathHelpers';
 
 const pathBase = `BanjoKazooie`;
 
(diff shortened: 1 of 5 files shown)


src/BanjoKazooie/tools/extractor.ts
```diff
@@ -59,16 +59,28 @@ interface CRG1File {
...
     return index;
 }
 
+function extractFile(fs: FS, fsfile: FSFile): CRG1File {
+    const fileBuffer = getFileBuffer(fs, fsfile);
+    const buffer = (fsfile.flags & 0x00010000) ? decompress(fileBuffer) : fileBuffer;
+    return { Data: buffer };
+}
+
+function decompressBigSlice(fs: FS, offset: number, end: number): ArrayBufferSlice {
```
PFedak: I couldn't tell if I was missing a more obvious way to do this. The decompress function was yelling about there not being enough room for the full data, as far as I could tell. I may still not fully understand the relationship between ArrayBuffers and the typed arrays, though.

src/BanjoKazooie/scenes.ts
```diff
@@ -120,7 +197,57 @@ class SceneDesc implements Viewer.SceneDesc {
...
+                    offs += otherSize * 0xc;
+                } else {
+                    offs += 12;
+                }
+                dataType = view.getInt8(offs++);
+            }
+        }
+    }
+
+    public async createScene(device: GfxDevice, context: SceneContext): Promise<Viewer.SceneGfx> {
```
PFedak: I tried to pattern this after the DataHolder you added for Pilotwings, so I'm assuming it was okay to make this async. I'm sure I messed up something subtle here though.

src/BanjoKazooie/tools/extractor.ts
```diff
@@ -126,6 +138,158 @@ function extractMap(fs: FS, name: string, sceneID: number): void {
...
+    // the uncompressed files are placed consecutively in RAM, so probably a better way to get these addresses
+    map.regions.push({ data: decompressBigSlice(fs, 0xF37F90, 0xF9CAE0), start: 0x80286F90 });
+    map.regions.push({ data: decompressBigSlice(fs, 0xF9CAE0, 0xFA3FD0), start: 0x80363590 });
+    map.regions.push({ data: decompressBigSlice(fs, 0xFC6F20, 0XFC8AFC), start: 0x803863F0 });
+    map.regions.push({ data: decompressBigSlice(fs, 0xFC8AFC, 0XFC9150), start: 0x8038D350 });
+
+    extractObjectLoadFromAssembly(fs, setupTable, map, 0x802c2c08);
+    extractObjectLoadFromAssembly(fs, setupTable, map, 0x8038c4f4);
+
+    const data = BYML.write({ ObjectSetupTable: setupTable }, BYML.FileType.CRG1);
```
PFedak: This seemed to just work, but please let me know if there are some best practices with this format I should be paying attention to.

magcius:
Merged manually.

src/BanjoKazooie/tools/extractor.ts
magcius: I'm unsure exactly what the problem is. I might take a look later.

src/BanjoKazooie/tools/extractor.ts
magcius: Yeah, it's basically a more efficient format for JSON, based on Nintendo's tooling.

src/BanjoKazooie/scenes.ts
magcius: Looks fine to me.

PR #170, opened 2019-11-11
Camera: Pan speed control

Mushman:
Panning with the keyboard uses camera speed while changing angle


magcius:
So the reason I didn't do this before is that the camera speed is often level-dependent, but how long it takes to 180 shouldn't be. I want to fix the former problem, but for now, a user can open up the F12 devtools and type `main.viewer.cameraController.keyAngleChangeVelSlow = 0.02` or something.

Mushman:
Fair enough. I might try to set up a separate control as we discussed.

PR #178, opened 2019-11-19
Convert project to Webpack 4

encounter:
Hi, I saw off-hand in the urde discord that you were looking for help trying out Webpack. Here's what I came up with that matches the existing build process as closely as possible.

Some notes on wasm:
- I tried making the AssemblyScript compilation a part of the Webpack build, but the existing loaders for it don't work well & are seemingly not maintained.
- Webpack v4 does include [async importing *.wasm files](https://github.com/webpack/webpack/tree/v4.41.2/examples/wasm-simple), but I did not try that out yet.

Instead, just using `CopyWebpackPlugin` to copy the existing wasm files to the dist root works for now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,27 +13,37 @@
     "@types/webgl2": "0.0.4",
     "@webgpu/types": "^0.0.15",
     "assemblyscript": "github:AssemblyScript/assemblyscript",
+    "clean-webpack-plugin": "^3.0.0",
+    "copy-webpack-plugin": "^5.0.5",
+    "file-loader": "^4.2.0",
+    "git-revision-webpack-plugin": "^3.0.4",
+    "html-webpack-plugin": "^3.2.0",
     "iconv-lite": "^0.4.24",
     "pngjs": "^3.4.0",
     "puppeteer": "^1.16.0",
-    "typescript": "^3.3.3333"
+    "raw-loader": "^3.1.0",
+    "ts-loader": "^6.2.1",
+    "typescript": "^3.3.3333",
+    "webpack": "^4.41.2",
+    "webpack-cli": "^3.3.10",
+    "webpack-dev-server": "^3.9.0",
+    "webpack-merge": "^4.2.2"
   },
   "dependencies": {
     "@sentry/browser": "^5.7.1",
+    "buffer": "^5.4.3",
     "crc-32": "^1.2.0",
     "gl-matrix": "^3.0.0",
     "librw": "^0.4.0",
     "pako": "^1.0.7",
-    "parcel-bundler": "^1.12.4",
     "reflect-metadata": "^0.1.13"
   },
   "scripts": {
-    "start": "parcel serve src/index.html",
-    "embed": "parcel serve src/embed.html",
+    "start": "webpack-dev-server --config webpack.dev.js",
     "asbuild:gx_tex": "npx asc gx_texture_as.ts --baseDir ./src/assembly -O3 --runtime none -b gx_texture_as.wasm -t gx_texture_as.wat",
     "asbuild:yaz0": "npx asc Yaz0_as.ts --baseDir ./src/assembly -O3 --runtime none --noAssert -b Yaz0_as.wasm -t Yaz0_as.wat",
     "asbuild": "yarn asbuild:gx_tex && yarn asbuild:yaz0 && node src/build_wat.js",
-    "build": "parcel build src/index.html src/embed.html && cp -f src/vendor/glslang/glslang.wasm src/vendor/basis_universal/basis_transcoder.wasm node_modules/librw/lib/librw.wasm dist/",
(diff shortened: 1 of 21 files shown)


package.json
```diff
@@ -13,27 +13,37 @@
...
+    "ts-loader": "^6.2.1",
+    "typescript": "^3.3.3333",
+    "webpack": "^4.41.2",
+    "webpack-cli": "^3.3.10",
+    "webpack-dev-server": "^3.9.0",
+    "webpack-merge": "^4.2.2"
   },
   "dependencies": {
     "@sentry/browser": "^5.7.1",
+    "buffer": "^5.4.3",
```
magcius: What is `buffer` used for?

encounter: That was from testing Webpack 5 beta, where many embedded polyfills have been removed. Some library was importing it. I'll remove it since this PR is back to Webpack 4.

magcius:
Would you also mind testing build performance and resulting binary bundle size?

encounter:
On my machine, comparing production builds:
Parcel:
```
✨  Built in 19.29s.

dist/main.abec6458.js               ⚠️  2.54 MB    18.51s
dist/embeds_main.2d8b492b.js        ⚠️  1.15 MB    18.16s
dist/main.abec6458.js.map           ⚠️  1.08 MB      66ms
dist/embeds_main.2d8b492b.js.map      372.5 KB      47ms
```
Webpack:
```
                              Asset        Size  Chunks                                Chunk Names
      embed-d17ad381d1acef192fe2.js    1.03 MiB       0  [emitted] [immutable]  [big]  embed
  embed-d17ad381d1acef192fe2.js.map    2.98 MiB       0  [emitted] [dev]               embed
       main-d7d119fdb2eef4236eba.js    2.27 MiB       1  [emitted] [immutable]  [big]  main
   main-d7d119fdb2eef4236eba.js.map    7.41 MiB       1  [emitted] [dev]               main

Done in 24.98s.
```
(Omitted non-js files for brevity)

magcius:
Alright. Let's try this out and see how it works. I'll revert if it goes bad.

package.json
magcius: It might have been Pako? I have very few dependencies, so I'm curious which one it was.

encounter:
Cool, deployment looks OK on my end.

webpack.prod.js
```diff
@@ -0,0 +1,13 @@
+const webpack = require('webpack');
+const merge = require('webpack-merge');
+const common = require('./webpack.common.js');
+
+module.exports = merge(common, {
+  mode: 'production',
+  devtool: 'source-map',
```
magcius: We can probably remove the source maps in production, since the original sources aren't uploaded to the server.

webpack.dev.js
```diff
@@ -0,0 +1,16 @@
+const webpack = require('webpack');
+const merge = require('webpack-merge');
+const common = require('./webpack.common.js');
+
+module.exports = merge(common, {
+  mode: 'development',
+  devtool: 'inline-source-map',
+  devServer: {
+    contentBase: './dist',
```
magcius: I forgot to ask, currently I have a hacky symlink solution to get `https://localhost:8080/data/` pointed to `./data/`, by way of symlink through `./dist/data/`. I assume the dev server can be slightly better about this?

webpack.prod.js
encounter: devtool: false will do that

webpack.dev.js
encounter: Sure, data and .htaccess could be moved into a static/ directory, then contentBase updated to point to static instead. Then the CleanWebpackPlugin configuration can be removed, using the default behavior of cleaning everything in dist. But then the deploy configuration would have to be updated to upload the contents of both dist and static, or a new Copy plugin added to prod only to copy static to dist. Though depending on how much data there is, the latter option may be slow.

magcius: I mean, we don't have to go that far. dist/ is used both for development and deployment, but differently:

* `dist/.htaccess` is used during deployment, but not development. (Does nothing on dev-server)
* `dist/data` is used during development, but not deployment. (The CDN backend is used for data instead)

Can we can come up with an approach that keeps `dist/` for the deployment case, and just mounts `./data` as `http://localhost:8080/data/`.

encounter: Nevermind, just realized the data dir contents aren't actually uploaded during website deploy, so you could skip the part about copying from static.

webpack.prod.js
encounter: That sped up the build by a good 15 seconds!

webpack.dev.js
encounter: Yeah, understood. I think we'd have to move ./data to something like ./static/data so contentBase can point to static. I don't see a devServer option to redirect a specific directory.

PR #180, opened 2019-11-21
ztp: Improve mirror render hack

Mushman:
Keeps spawned objects in the correct place, despite not being set yet.


src/j3d/ztp_scenes.ts
```diff
@@ -96,8 +97,9 @@ class TwilightPrincessRenderer implements Viewer.SceneGfx {
     }
 
     private setMirrored(mirror: boolean): void {
+        const negScaleMatrix = mat4.fromValues(-1,0,0,0,0,1,0,0,0,0,1,0,0,0,0,1);
```
magcius: I wonder if we should have a `computeModelMatrixS(-1, 0, 0)` in MathHelpers.ts for this. I really should turn that into a fuller matrix utility library...

Mushman: Not a bad idea. I'm also adding mirroring to Mario Kart Double Dash, so I've taken the time to make a new commit with this helper function, as well as updating the Mario Kart Wii implementation too.

PR #186, opened 2019-12-04
JSR: Add viewer for Dreamcast PowerVR textures

x1nixmzeng:
This code allows drop-and-drop support for the PVR texture files used in Jet Set Radio (Dreamcast) - and other Dreamcast titles with the same loose format.

I took this task from the list of 'games to get start on' on the Discord server, but it was also my first exposure to typescript and I wanted to contribute!

There is also a start on the AFS container, which as well as textures, contains loc text and possibly model data.


src/Common/DC/PVRT.ts
```diff
@@ -0,0 +1,477 @@
...
+        hostAccessPass.uploadTextureData(gfxTexture, 0, textureEntry.levels.map((level) => level.data));
+        device.submitPass(hostAccessPass);
+
+        const viewerTexture: Viewer.Texture = textureToCanvas(textureEntry);
+
+        return { gfxTexture, viewerTexture };
+    }
+}
+
+class Untwiddle {
```
magcius: You shouldn't need the table, it won't be a big performance boost. Just leave untwiddleValue / getUntwiddledTexelPosition as functions.

src/Common/DC/PVRT.ts
```diff
@@ -0,0 +1,477 @@
...
+        else {
+            pos = this.twittleTable[y] | this.twittleTable[x] << 1;
+        }
+        
+        return pos;
+    }
+}
+
+function unpackTexelToRGBA(srcTexel: number, srcFormat: PVRTFormat, dst: Uint8Array, dstOffs: number): void
+{
```
magcius: Style nit: Braces go on same line (please fix this in the rest of the file)

src/Common/DC/PVRT.ts
```diff
@@ -0,0 +1,477 @@
...
+    switch( srcFormat )
```
magcius: I tend to prefer if statements over switch statements these days.

src/Common/DC/PVRT.ts
```diff
@@ -0,0 +1,477 @@
...
+
+    return params;
+}
+
+function decideLevels(width: number, height: number, params: UnpackedParams): UnpackedLevel[] {
+    let levels: UnpackedLevel[] = [];
+
+    let srcOffset = 0;
+
+    if (params.vqCompressed == true) {
```
magcius: `if (params.vqCompressed)`

PR #190, opened 2019-12-18
Add basic touch controls

beholdnec:
This PR adds basic (crude) touch controls, in order to make noclip usable on mobile devices.

The following touch controls are implemented:
- [x] Single-touch and drag to perform the equivalent of mouse dragging.
- [x] Double-touch and pinch (or spread apart) to move forward or back in FPS mode.

The following features would be desirable, but are not implemented by this PR:
- [ ] A button to open or close the UI. This is currently difficult on touch devices in portrait mode.
- [ ] Specialized touch controls for orbit and orthographic cameras.
- [ ] Speed controls

### DIFF
--- a/src/InputManager.ts
+++ b/src/InputManager.ts
@@ -110,6 +110,16 @@ export default class InputManager {
         return this.dy;
     }
 
+    public getTouchDeltaX(): number {
+        // XXX: In non-pinch mode, touch deltas are turned into mouse deltas.
+        return this.touchGesture == TouchGesture.Pinch ? this.dTouchX : 0;
+    }
+
+    public getTouchDeltaY(): number {
+        // XXX: In non-pinch mode, touch deltas are turned into mouse deltas.
+        return this.touchGesture == TouchGesture.Pinch ? this.dTouchY : 0;
+    }
+
     public getPinchDeltaDist(): number {
         return this.touchGesture == TouchGesture.Pinch ? this.dPinchDist : 0;
     }
@@ -130,6 +140,8 @@ export default class InputManager {
         this.dx = 0;
         this.dy = 0;
         this.dz = 0;
+        this.dTouchX = 0;
+        this.dTouchY = 0;
         this.dPinchDist = 0;
 
         // Go through and mark all keys as non-event-triggered.
--- a/src/InputManager.ts
+++ b/src/InputManager.ts
@@ -43,6 +43,10 @@ export default class InputManager {
     private usePointerLock: boolean = true;
     public isInteractive: boolean = true;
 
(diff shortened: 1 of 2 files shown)


src/InputManager.ts
```diff
@@ -159,6 +178,49 @@ export default class InputManager {
...
+        if (e.touches.length == 1) {
+            this.isSingleTouching = true;
+            this.isDoubleTouching = false;
+            this.onMotion(e.touches[0].clientX - this.prevSingleTouchX, e.touches[0].clientY - this.prevSingleTouchY);
+            this.prevSingleTouchX = e.touches[0].clientX;
+            this.prevSingleTouchY = e.touches[0].clientY;
+        } else if (e.touches.length == 2) {
+            this.isDoubleTouching = true;
+            this.isSingleTouching = false;
+            const newDist = vec2.dist([e.touches[0].clientX, e.touches[0].clientY], [e.touches[1].clientX, e.touches[1].clientY]);
```
magcius: `Math.hypot(e.touches[0].clientX - e.touches[1].clientX, e.touches[0].clientY - e.touches[1].clientY)` might be a bit easier to read.

src/InputManager.ts
```diff
@@ -100,13 +118,14 @@ export default class InputManager {
     }
 
     public isDragging(): boolean {
-        return GlobalGrabManager.hasGrabListener(this);
+        return this.isSingleTouching || GlobalGrabManager.hasGrabListener(this);
```
magcius: I feel like isDragging should be true if the number of touches is non-zero...

beholdnec: Perhaps so. The reason I use isSingleTouching is that I wanted single-touching to map fully onto mouse-dragging, whereas double-touching does not map onto mouse-dragging, since there is no mouse control that moves the camera forward and back. It's a matter of semantics.

beholdnec:
Update: This PR now adds the ability to move the camera horizontally and vertically while pinching, similar to SketchFab.

src/InputManager.ts
magcius: I think if we were to add some hypothetical mouse control to move forwards and backwards, it would be considered dragging, though. Things that isDragging controls:

* Whether the camera dynamics friction is in the "dragging" mode for Orbit/Ortho modes (if you put your hand on a ceiling fan to stop it, it stops faster rather than spinning down naturally, that thing), which I feel like any number of touches should do.
* Whether to hide the new bottom-bar UI, on the premise that if you're interacting with the scene, you're not interacting with the bottom-bar. Similarly, I feel like isDragging should be true here for double-dragging.

PR #192, opened 2019-12-20
GX: Material Display List parsing supports lighting registers

themikelester:
The XF registers related to lighting and/or vertex colors are read by parseMaterialEntry(). This was previously a TODO in WindWaker/Actors.ts. That function now also determines cull state and TEV/Indirect/TexGen counts internally. This makes DL -> GXMaterial conversion much simpler. 

I've also used this to fix the lighting for flowers in WindWaker. Two additional unrelated problems are fixed as well:
- The color arrays for white and bessou flowers cannot be extracted from the .dol. I've hardcoded them as they're only 8 bytes.
- Flowers do not set Color0, they rely on it being black

GxMaterial currently has no way of storing lighting material/ambient color register values. This isn't inherently a problem, but I've added a warning to parseMaterialEntry in case a display list is found that uses them. Ideally there'd be some kind of GxMaterial++ structure that encapsulates all the data that a material display list contains. As far as I know that is textures and register values. 

## Before
Screen Shot 2019-12-20 at 4 56 40 AM

## After
Screen Shot 2019-12-20 at 4 54 05 AM

## Original
Screen Shot 2019-12-14 at 11 35 09 AM


src/rres/brres.ts
```diff
@@ -749,8 +777,7 @@ function parseMDL0_MaterialEntry(buffer: ArrayBufferSlice, version: number): MDL
     parseMDL0_TevEntry(buffer.subarray(tevOffs), r, numTevs);
 
     // Now combine the whole thing.
-    const gxMaterial = parseMaterialEntry(r, index, name, numTexGens, numTevs, numInds);
-    gxMaterial.cullMode = cullMode;
+    const gxMaterial = parseMaterialEntry(r, index, name);
```
magcius: So BRRES's parseMaterialEntry is designed for BRRES, where I know what will and won't appear in the material entry list. Cull Mode is not part of the material entry DL, so this will break cull mode for BRRES.

I don't think GenMode is part of this register collection either, hence why this was separate. One option here would be to set the registers from the fields above, but that might be annoying to construct.

Basically, please ensure that you don't break BRRES through this PR.

src/j3d/WindWaker/Actors.ts
```diff
@@ -474,7 +450,6 @@ export class FlowerObjectRenderer implements ObjectRenderer {
             return;
 
         materialParams.m_TextureMapping[0].copy(this.flowerData.textureMapping);
-        colorCopy(materialParams.u_Color[ColorKind.C0], this.c0);
```
magcius: Probably want to remove `this.c0` as well.

themikelester:
Closing in favor of #193

PR #193, opened 2019-12-21
GX: Generalize material display list parsing out of brres and into gx_displaylist/gx_material

themikelester:
Based on the discussion from #192 

The BRRES file format was previously the only user of serialized material display lists. Wind Waker also makes use of them, but reads more registers. 

This PR does a few things:
- Moves `displayListRegistersRun()` (which sets registers based on display list commands) to gx_displaylist.ts. 
- Moves register -> GXMaterial parsing logic out of brres.ts and into gx_material.ts. 
- Added `parseMaterial()` (from display list registers) which also parses lighting channels
- Simplified Wind Waker grass actors to use the new material parsing functionality

Note: I wasn't able to test BRRES usage, but there were no logic changes


magcius:
> Note: I wasn't able to test BRRES usage, but there were no logic changes

FYI, you can change DataFetcher.ts to use the live data store if you want to test stuff without having the files locally.

themikelester:
That's exactly what I wanted to know! Thanks

PR #194, opened 2019-12-22
Wind Waker: A bit more control over time of day handling

themikelester:
_This one's been sitting around in my fork for a while. What do you think?_

ToD starts at DAY by default, with a 60 second day cycle (10 seconds per phase).


magcius:
I'm fine with it. I used to have it be a separate option box on the left, then I hooked it up to time but had a slider to let you scrub through time, and then I removed the time scrubber. This is a bit of a sore point right now, I agree.

PR #196, opened 2019-12-24
GX: Added displayListToString() to help debugging

themikelester:
This basically outputs a string like `Setting BP GEN_MODE_ID to 0x12345678` for each command in a *material* display list. Currently it will break if there are shape commands in there.

This PR also fixes a bug that would prevent modification of the BP SS_MASK register. It needs to remain set until the next command. 

I'm currently using this for tree debugging.


src/gx/gx_displaylist.ts
```diff
@@ -1045,7 +1045,8 @@ export class DisplayListRegisters {
         // Retrieve existing value, overwrite w/ mask.
         const regValue = (this.bp[regAddr] & ~regWMask) | (regBag & regWMask);
         // The mask resets after use.
-        this.bp[GX.BPRegister.SS_MASK] = 0x00FFFFFF;
+        if (regAddr != GX.BPRegister.SS_MASK)
```
magcius: !==

src/gx/gx_enum.ts
```diff
@@ -642,6 +680,104 @@ export const enum CPRegister {
     VAT_C_ID      = 0x90,
 }
 
+export const XFRegisterStrings = new Map([
+    [XFRegister.XF_INVTXSPEC_ID    , 'XF_INVTXSPEC_ID'],
```
magcius: If you change the const enum's to enum's (and actually, const enum is not working right now for fun reasons), TypeScript will do the inverse mapping for you, you can just do `XFRegister[reg]`.

src/gx/gx_enum.ts
```diff
@@ -588,25 +588,47 @@ export const enum BPRegister {
     // SetTevOrder
     RAS1_TREF_0_ID     = 0x28,
 
+    // Tex offsets
+    SU_SSIZE_I0_ID = 0x30,
```
magcius: FYI, I do not handle SSIZE "correctly" right now, in that I don't allow SSIZEs that are different from the texture size. If this happens in practice though, I'll be shocked.

themikelester: No worries, these are mostly just so that I can catch anyone trying to set these and other unusual/unimplemented things

src/gx/gx_enum.ts
themikelester: So good to know!

themikelester:
Updated to fix the comments

PR #197, opened 2019-12-25
Wind Waker: Add optional normal output to DZB `raycast()`

themikelester:
This is needed for tree shadows (they orient themselves based on the ground normal). I'm definitely up for doing this another way, this just seemed to be the most straightforward.

### DIFF
--- a/src/j3d/WindWaker/DZB.ts
+++ b/src/j3d/WindWaker/DZB.ts
@@ -122,8 +122,10 @@ const pt0 = vec3.create();
 const pt1 = vec3.create();
 const pt2 = vec3.create();
 const bary = vec3.create();
-export function raycast(closestHit: vec3, dzb: DZB, origin: vec3, direction: vec3): boolean {
+export function raycast(closestHit: vec3, dzb: DZB, origin: vec3, direction: vec3, outNormal?: vec3): boolean {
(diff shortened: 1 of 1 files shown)


src/j3d/WindWaker/DZB.ts
```diff
@@ -153,11 +155,27 @@ export function raycast(closestHit: vec3, dzb: DZB, origin: vec3, direction: vec
...
+
+        if (outNormal) {
+            vec3.set(pt0, dzb.vertexData[ci0*3+0], dzb.vertexData[ci0*3+1], dzb.vertexData[ci0*3+2]);
+            vec3.set(pt1, dzb.vertexData[ci1*3+0], dzb.vertexData[ci1*3+1], dzb.vertexData[ci1*3+2]);
+            vec3.set(pt2, dzb.vertexData[ci2*3+0], dzb.vertexData[ci2*3+1], dzb.vertexData[ci2*3+2]);
+
+            const v0 = vec3.sub(pdir, pt1, pt0);
+            const v1 = vec3.sub(bary, pt2, pt0);
+            vec3.cross(outNormal, v0, v1);
+            vec3.normalize(outNormal, outNormal);
```
magcius: I think the DZB file stores the surface normal explicitly, but this is fine for now.

PR #199, opened 2019-12-25
Wind Waker: Grass, Flowers, and Trees! 

themikelester:
Screen Shot 2019-12-25 at 4 56 12 PM

Fully correct (as far as I can tell) animation and "lighting" for all grass, flower, and tree actors. The structure in this PR mimics that of the game. There are three separate "packet" classes which manage trees/grass/flowers. There is a single actor rel named 'da_grass' which spawns some number of elements in one of these packets.

This PR is still a bit sloppy. Some things that could use attention:
- The draw() functions in each of the three packets. Am I doing this optimally? I don't feel very comfortable with the renderInst pipeline yet.
- ~The new kyanko stuff in zww_scenes. I wanted full kyanko colors for each room, because that's how the game does it. Now that I understand how the colors are used, this isn't really necessary. I'll probably remove the new kyanko stuff and just assume a single room.~ EDIT: I've reverted it back to the original system, with the small change that `dstColors` is now public
- There are two new "contexts":
  - The Actor context basically mimics the game's Actor class.
  - WindWakerRenderer is now being treated like the game's global namespace. I've added more properties that subsystems like Grass need. 
- The three systems do not currently observe the 'disable textures/vert colors/objects' render hacks. I still need to implement that.
- ~Oops, there's also no frustum or distance culling~ EDIT: Added distance culling

Nits and larger changes welcome, this is still pretty rough.


themikelester:
Oh I should also note, zww_extractor has changed, so the new extra.crg1 file needs to be updated

src/j3d/WindWaker/Actors.ts
```diff
@@ -25,6 +25,16 @@ import { GfxRenderCache } from '../../gfx/render/GfxRenderCache';
 import { GfxRenderInstManager } from '../../gfx/render/GfxRenderer';
 import { BTIData, BTI_Texture } from '../../Common/JSYSTEM/JUTTexture';
 
+export interface Actor {
+    name: string,
```
magcius: Code nit: end lines with semicolons in interface blocks, not commas.

src/j3d/WindWaker/Grass.ts
```diff
@@ -0,0 +1,1229 @@
...
+            this.anims[i] = {
+                active: true,
+                rotationX: 0,
+                rotationY: i * dy,
+                matrix: mat4.create(),
+            }
+        }
+    }
+
+    newData(pos: vec3, type: FlowerType, roomIdx: number, itemIdx: number): FlowerData {
```
magcius: Code nit: Please use public/protected/private as appropriate.

src/j3d/WindWaker/Grass.ts
```diff
@@ -0,0 +1,1229 @@
...
+            mat4.rotateX(anim.trunkMtx, anim.trunkMtx, anim.trunkRotationX);
+            mat4.rotateY(anim.trunkMtx, anim.trunkMtx, uShortTo2PI(anim.initialRotationShort) - anim.trunkFallYaw);
+        }
+
+        for (let i = 0; i < kMaxFlowerDatas; i++) {
+            const data = this.datas[i];
+            if (!data) continue;
+
+            // Perform ground checks for some limited number of data
+            if (data.flags & TreeFlags.needsGroundCheck && groundChecksThisFrame < kMaxGroundChecksPerFrame) {
```
magcius: Put parens around bit checks. Precedence rules means it likely doesn't do what you want.

themikelester:
Fixed some of the nits, and a few of mine

LagoLunatic:
Thought I'd mention there's an issue when loading "The Great Sea" scene where not all the foliage appears even though it does in the individual islands:
image

"Failed to allocate data" appears in the console a bunch. I think it's because of the 1500 kMaxGrassDatas limit. The first few islands have foliage, but after the limit is hit the later islands don't.

LagoLunatic:
Oh, and there's a bug on the private oasis where both white and pink flowers get replaced by the bessou flowers:
image
Ingame, only pink flowers get replaced:
GZLE99_2019-12-27_18-22-51

themikelester:
Thanks Lago! Good catches. I'll get to these soon.

themikelester:
The Bessou flowers are fixed:
Screen Shot 2019-12-27 at 4 59 08 PM

PR #203, opened 2019-12-27
BFBB: Improved PIPT and lighting support

pslehisl:
PIPT settings are supported more accurately, allowing for per-object alpha testing and fog/light filtering. Basic gouraud shading is implemented, looks decent in some levels, bad in others. I added a render hacks panel for toggling lighting on/off.

### DIFF
--- a/src/SpongeBobBFBB/program.glsl
+++ b/src/SpongeBobBFBB/program.glsl
@@ -52,7 +52,7 @@ void main() {
     v_Color = a_Color;
     v_TexCoord = a_TexCoord;
 
-    vec3 t_Normal = normalize(Mul(transpose(inverse(_Mat4x4(u_ModelMatrix))), vec4(a_Normal, 1.0)).xyz);
+    vec3 t_Normal = normalize(Mul(_Mat4x4(u_ModelMatrix), vec4(a_Normal, 0.0)).xyz);
 
 #ifdef USE_LIGHTING
     if (USE_LIGHTING == 1) {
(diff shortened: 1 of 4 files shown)


src/SpongeBobBFBB/program.glsl
```diff
@@ -52,28 +52,28 @@ void main() {
     v_Color = a_Color;
     v_TexCoord = a_TexCoord;
 
-    //vec4 t_Normal = Mul(transpose(inverse(_Mat4x4(u_ModelMatrix))), vec4(a_Normal, 1.0));
+   // vec4 t_Normal = Mul(transpose(inverse(_Mat4x4(u_ModelMatrix))), vec4(a_Normal, 1.0));
     //vec4 t_Normal = vec4(a_Normal, 1.0);
```
magcius: You probably want `vec3 t_Normal = Mul(_Mat4x4(u_ModelMatrix, vec4(a_Normal, 0.0))).xyz;`

pslehisl: The inverse transpose matrix is needed for non-uniformly scaled models. What I did wrong was use the euler angles from each light's matrix, instead of the forward vector, as the light's direction. This is fixed in my latest commit.

magcius: I'm aware, but that's relatively rare, and the artifacts are not major, that you can usually get away with not computing the inverse transpose. Note that inverse/transpose won't work for systems that require the row-major workaround, and computing the matrix inverse is expensive, so it should be done on the CPU if absolutely necessary.

The big important change change here is the fourth component being 0.0, since normals should not get translation.

pslehisl: Fair enough.

PR #211, opened 2020-01-06
BJ: Compile new fragment shader per combine parameters

beholdnec:
Currently, the Banjo-Kazooie renderer uses dynamic array indexing to get combine parameters. With this patch, a new fragment shader is compiled for each set of combine parameters. This greatly improves performance on one of my weaker machines, and is potentially compatible with more browsers.

### DIFF
--- a/src/BanjoKazooie/render.ts
+++ b/src/BanjoKazooie/render.ts
@@ -44,7 +44,6 @@ layout(row_major, std140) uniform ub_DrawParams {
 };
 
 uniform ub_CombineParameters {
-    vec4 u_Params;
     vec4 u_PrimColor;
     vec4 u_EnvColor;
 };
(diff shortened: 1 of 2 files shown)


magcius:
I am surprised the dynamic indexing is so slow, huh.

I believe the Pilotwings 64 code also uses the same F3DEX shader, if not a very similar one? And there's also Paper Mario 64. It might be nice to deduplicate these eventually, but PM64 will take some work.

src/BanjoKazooie/render.ts
```diff
@@ -151,6 +151,37 @@ void main() {
...
+            't_Tex0.aaa', 't_Tex1.aaa', 'u_PrimColor.aaa', 'v_Color.aaa',
+            'u_EnvColor.aaa', 't_Zero.rgb' /* LOD */, 't_Zero.rgb' /* prim LOD */, 't_Zero.rgb'
+        ];
+
+        const alphaInputs: string[] = [
+            'combAlpha', 't_Tex0', 't_Tex1', 'u_PrimColor.a',
+            'v_Color.a', 'u_EnvColor.a', '1.0', '0.0'
+        ];
+
+        function unpackParams(params: number): {x: number, y: number, z: number, w: number} {
```
magcius: The parameter packing was done for the sake of shaders, so we could pack all four combines into one vector. It doesn't really need to be done this way if we have all of JS to work with. Not too important, but something to keep in mind for future cleanups.

magcius:
FYI, If this works with Pilotwings 64, go ahead and merge it. I believe you should have permissions to press the merge button.

beholdnec:
Pilotwings 64 fails with this patch. I didn't realize they shared code. Don't merge.

magcius:
It shouldn't be too much work to fix Pilotwings 64 -- I can do it later today, even.

beholdnec:
Pilotwings 64 now works again, but please review Pilotwings64/Scenes.ts, line 1688 . chosenCombine is undefined and so no combine info is available. I'm not sure what's going on there, but the results look ok to me.

magcius:
@PFedak I think you were the one to touch this last?

PR #214, opened 2020-01-16
Implement N64 Ocarina of Time

beholdnec:
Here is a PR that brings back support for the Nintendo 64 version of The Legend of Zelda: Ocarina of Time. The .zelview files already exist on the server, so merging this PR is all that is needed to support this game.

This code is probably good enough to at least be considered "experimental". Multitexturing is much improved from when this game was dropped from noclip.

The following features are not implemented yet:
- Depth sorting of translucent objects


src/zelview/zelview0.ts
```diff
@@ -0,0 +1,458 @@
...
+            const file = rom.lookupFile(pStart);
+            const room = readRoom(file);
+            room.filename = file.filename;
+            rooms.push(room);
+            roomTableAddr += 8;
+        }
+        return rooms;
+    }
+
+    // function loadImage(gl: WebGL2RenderingContext, src: string) {
```
magcius: Could you remove these older functions?

beholdnec:
I've updated the PR to address feedback. Also, lighting in outdoor scenes is implemented.

magcius:
So one thing I remember from the earlier zelview implementation is that at the top of Inside the Deku Tree, the webbing became solid, and when I investigated, I remembered that the CC flags were junk. Could you double-check that?

Also, could you move to Experimental for the time being?

PR #232, opened 2020-02-18
map relabeling

HALLEYSHURA:
tweaked map labels for SFA and okami


beholdnec:
Thumbs up for SFA names. It's been a while since I played the game so I couldn't remember all the area names.

Does "Box Cage" (which you named "CloudRunner Fortress 2?") actually appear in the game? That map contains the only instances of swastika boxes I can find in the final game. The rest of the boxes were given a 3-pointed emblem before the game was released.

PR #236, opened 2020-02-21
SFA/DP ancient maps labeled

HALLEYSHURA:
most of the ancient maps are DP versions of SFA maps and are labeled as such. the cloudrunner fortress levels are closer to retail SFA, so are labeled as early SFA. question marks are on ones i'm unsure of; there's nothing in SFA or existing DP footage that resembles the pit or mine rooms, so those are marked as "unknown".


magcius:
Please don't use acronyms like DP/SFA without explanation.

beholdnec:
I don't think there's a need to label SFA and DP. They're old and new versions of the same game.

HALLEYSHURA:
ah, alright. you're free to remove the acronyms; i only thought it was really necessary for the cloudrunner fortress maps since they look pretty different from what's available of the Dinosaur Planet version:
https://www.unseen64.net/wp-content/uploads/2008/04/dinosaurplanet865.JPG
https://www.models-resource.com/gamecube/starfoxadventures/model/23517/ (the world map model for cloudrunner fortress resembles the dinoplanet version as opposed to the final game; seems to be a lower-poly version of the level)

the thorntail hollow map also has a bit more detail that i didn't think would be possible on N64, but that's just speculation on my end. regardless, i'll remove the acronyms unless you guys wanna do that o:

magcius:
Thanks for the quick fixes!

PR #239, opened 2020-02-26
WebXR implementation

### LINKED ISSUE #106: Any plans to add WebVR/WebXR support?
It would be incredible to see some of these levels in VR, especially without worrying about issues like frustum culling or getting real emulators to support VR.

fordacious:
Closes #106 

An implementation of WebXR. Attempted to implement it as completely as possible with as little disruption as possible. Still some rough edges which are documented in the WebXR.ts file.

Thanks to @MelonSpeedruns for help on the UI and Interaction code and testing.
Thanks to Rainwolf and Hillexed for testing help.


package.json
```diff
@@ -41,6 +41,7 @@
     "reflect-metadata": "^0.1.13"
   },
   "scripts": {
+    "zww_extract": "tsc ./src/WindWaker/tools/zww_extractor.ts --target es5 --downlevelIteration --outDir ./src/WindWaker/tools/build && cd ./src/WindWaker/tools && node ./build/WindWaker/tools/zww_extractor.js",
```
fordacious: not sure how this got in here.

src/main.ts
```diff
@@ -195,6 +195,9 @@ class Main {
 
         this.canvas = document.createElement('canvas');
 
+        this.canvas.width = 2200;
```
fordacious: remove

magcius:
Please squash into a single commit.

src/gfx/platform/GfxPlatformWebGL2.ts
```diff
@@ -1652,7 +1653,16 @@ void main() {
...
+        // TODO WebXR: stopgap to prevent overwriting of multiple viewports.
+        const currentScissor = gl.getParameter(gl.SCISSOR_BOX);
+        const currentScissorEnabled = gl.getParameter(gl.SCISSOR_TEST);
+        if (!!this.webXRContext && !!this.webXRContext.getCurrentViewport()) {
+            const vp: XrViewPort = this.webXRContext.getCurrentViewport();
+            this.setScissor(vp.x, vp.y, vp.width, vp.height);
+        } else {
+            gl.disable(gl.SCISSOR_TEST);
+        }
+
```
magcius: No. When specifying beginRenderPass with a disposition of CLEAR, you need to clear the entire target. Render to a smaller target and then composite into the big buffer if necessary.

src/viewer.ts
```diff
@@ -228,13 +307,10 @@ export class Viewer {
         this.inputManager.afterFrame();
 
         const deltaTime = dt * this.sceneTimeScale;
-        this.viewerRenderInput.deltaTime += deltaTime;
+        this.viewerRenderInput.deltaTime = deltaTime;
```
magcius: This shouldn't be necessary.

src/viewer.ts
```diff
@@ -106,7 +109,7 @@ export class Viewer {
     private clearScene: ClearScene = new ClearScene();
     private resolveRenderPassDescriptor = makeEmptyRenderPassDescriptor();
 
-    constructor(private gfxSwapChain: GfxSwapChain, public canvas: HTMLCanvasElement) {
+    constructor(private gfxSwapChain: GfxSwapChain, public webXRContext: WebXRContext, public canvas: HTMLCanvasElement) {
```
magcius: Viewer should not have any XR-specific code. Drive it from main.

src/viewer.ts
```diff
@@ -121,6 +124,12 @@ export class Viewer {
             viewport: this.viewport,
             onscreenTexture: null!,
         };
+
+        if (this.webXRContext) {
+            this.webXRContext.onSessionStarted = this.onWebXRStarted.bind(this);
+            this.webXRContext.onSessionEnded = this.onWebXREnded.bind(this);
+            this.webXRContext.onRender = this.render.bind(this, true);
```
magcius: render() *needs* to be called from updateDT. This is not negotiable.

src/WebXR.ts
```diff
@@ -0,0 +1,108 @@
...
+    public resetCurrentViewport(): void {
+        this.currentViewport = null;
+    }
+
+    public IsRunning(): boolean {
+        return this.isRunning;
+    }
+
+    public start() {
+        var _this = this;
```
magcius: Unused.

src/WebXR.ts
```diff
@@ -0,0 +1,108 @@
...
+        return navigator.xr.requestSession('immersive-vr', {
```
magcius: Please use async methods.

src/viewer.ts
```diff
@@ -139,24 +148,23 @@ export class Viewer {
...
-        this.gfxSwapChain.configureSwapChain(this.canvas.width, this.canvas.height);
-
-        this.renderStatisticsTracker.beginFrame();
-
-        resetGfxDebugGroup(this.debugGroup);
-        this.gfxDevice.pushDebugGroup(this.debugGroup);
+    private onWebXRStarted (): void {
+        var translation = vec3.create();
+        mat4.getTranslation(translation, this.camera.worldMatrix);
+        this.xrCameraController.setOffset(translation);
```
magcius: `mat4.getTranslation(this.xrCameraController.offset, this.camera.worldMatrix);`

Don't create GC garbage unless you can't help it.

src/viewer.ts
```diff
@@ -139,24 +148,23 @@ export class Viewer {
...
+    }
 
-        this.viewerRenderInput.onscreenTexture = this.gfxSwapChain.getOnscreenTexture();
+    private onWebXREnded (): void {
+        resizeCanvas(this.canvas, window.innerWidth, window.innerHeight, window.devicePixelRatio);
```
magcius: ???

src/viewer.ts
```diff
@@ -174,8 +182,79 @@ export class Viewer {
...
+            this.viewerRenderInput.backbufferHeight = this.canvas.height;
+            this.gfxSwapChain.configureSwapChain(this.canvas.width, this.canvas.height);
+            this.viewerRenderInput.onscreenTexture = this.gfxSwapChain.getOnscreenTexture();
+            
+            this.viewerRenderInput.viewport.x = 0;
+            this.viewerRenderInput.viewport.y = 0;
+            this.viewerRenderInput.viewport.w = 1;
+            this.viewerRenderInput.viewport.h = 1;
+
+            this.renderViewport(this.webXRContext.IsRunning());
```
magcius: Why does the non-XR codepath need to be modified?

src/ui.ts
```diff
@@ -1629,6 +1630,59 @@ class ViewerSettings extends Panel {
...
+    private EnableXRCheckBox: Checkbox;
+    private scaleSlider: Slider;
+    private webXRContext: WebXRContext;
+
+    constructor(private ui: UI, private viewer: Viewer.Viewer) {
+        super();
+
+        this.webXRContext = viewer.webXRContext;
+
+        this.setTitle(VR_ICON, 'XR Settings');
```
fordacious: Use VR in UI elements

src/viewer.ts
fordacious: this should reset if no WASD camera offset is available.

src/viewer.ts
MelonSpeedruns: This resizes the canvas back to the window size instead of the VR headset's screen size.

package.json
MelonSpeedruns: My bad. I used it to extract the wind waker assets and forgot to remove it...

src/viewer.ts
fordacious: Can you elaborate on this. WebXR uses a different requestAnimationFrame cadence so this is possible but tricky.

fordacious: Currently updates happen as normal, and the requestAnimationFrame for WebXR only renders.

src/viewer.ts
fordacious: This prevents rendering with a delta time of 0 (since it is set to 0 at the end anyway, the effect is the same).

src/viewer.ts
magcius: Call updateDT from within WebXR's requestAnimationFrame.

src/viewer.ts
magcius: Why are you resizing the canvas to begin with

src/viewer.ts
magcius: If you call updateDT() to render, as you are supposed to do, that should not matter.

src/gfx/platform/GfxPlatformWebGL2.ts
fordacious: Can you expand on how to go about this in your architecture?

src/viewer.ts
fordacious: Rendering to it as a stopgap for the time being

src/viewer.ts
fordacious: Was hoping to avoid tying the two request animation frames together but can do.

magcius: You can disable the main requestAnimationFrame and switch wholesale over to the XRSession's.

fordacious: as in that is what I was hoping to avoid :P

src/viewer.ts
magcius: I thought you had to render to an XrLayer for it to work.

fordacious: ah you're right. Sorry I was operating off faulty assumptions

src/gfx/platform/GfxPlatformWebGL2.ts
magcius: Set a custom backbufferWidth/Height so the game renders to a small area, and then blit it to the onscreen. Might have to change the API for GfxSwapChain::present to allow a custom viewport there.

src/Camera.ts
```diff
@@ -437,6 +449,111 @@ export class FPSCameraController implements CameraController {
...
+                if (gamepad.axes.length >= 3 && gamepad.buttons.length >= 1) {
+                    keyMovement[0] = gamepad.axes[2] * cameraMoveSpeed;
+                    keyMovement[1] = (gamepad.buttons[0].value - gamepad.buttons[1].value) * cameraMoveSpeed;
+                    keyMovement[2] = gamepad.axes[3] * cameraMoveSpeed;
+                }
+            }
+        }
+               
+        if (!vec3.exactEquals(keyMovement, vec3Zero)) {            
+            var viewMovementSpace = webXRContext.xrViewSpace.getOffsetReferenceSpace(
```
magcius: Don't use var.

src/Camera.ts
```diff
@@ -437,6 +449,111 @@ export class FPSCameraController implements CameraController {
...
+                new XRRigidTransform(
+                    new DOMPointReadOnly(keyMovement[0], 0, keyMovement[2], 1), {x:0, y:0, z:1.0, w: 1.0}));
+            var pose = webXRContext.currentFrame.getPose(viewMovementSpace, webXRContext.xrLocalSpace);
+
+            if (pose) {
+                var movementTranslation = vec3.create();
```
magcius: Use a scratch vector if necessary. This would probably just be cleaner as: ```
this.offset[0] += pose.transform.position.x;
...
```

File: src/Camera.ts [hunk 15]
```diff
@@ -437,6 +449,111 @@ export class FPSCameraController implements CameraController {
...
+                movementTranslation[0] = pose.transform.position.x;
+                movementTranslation[1] = keyMovement[1];
+                movementTranslation[2] = pose.transform.position.z;
+                vec3.add(this.offset, this.offset, movementTranslation);
+            }
+
+            updated = true;
+        }
+
+        assert(this.cameras.length == webXRContext.views.length);
```
magcius: Style nit: triple equals.

src/Camera.ts
```diff
@@ -437,6 +449,111 @@ export class FPSCameraController implements CameraController {
...
+        for (var i = 0; i < this.cameras.length; i++) {
+            const camera : Camera = this.cameras[i];
```
magcius: style nit: no space after type declarations, and I prefer to omit in most cases. `const camera = this.cameras[i];`

src/Camera.ts
```diff
@@ -437,6 +449,111 @@ export class FPSCameraController implements CameraController {
...
+            let xrView = webXRContext.views[i];
```
magcius: Use `const` when the binding does not need to be mutable.

src/WebXR.ts
```diff
@@ -0,0 +1,72 @@
...
+        this.currentFrame = frame;
+
+        if (pose) {
+            this.views = pose.views;
+        }
+
+        this.onFrame(time);
+
+        session.requestAnimationFrame(this.onXRFrame.bind(this));
+      }
```
magcius: Indentation is weird here.

src/WebXR.ts
```diff
@@ -0,0 +1,72 @@
...
+    public async start() {
+        this.xrSession = await navigator.xr.requestSession('immersive-vr', {
+            requiredFeatures: [],
+            optionalFeatures: ['viewer', 'local']
+        });
+
+        this.xrViewSpace = await this.xrSession.requestReferenceSpace('viewer');
+        this.xrLocalSpace = await this.xrSession.requestReferenceSpace('local');
+
+        let glLayer = new XRWebGLLayer(this.xrSession, this.renderingContext);
```
magcius: Preference would be to have this as an API on GfxPlatform: `const layer = gfxDevice.createWebXRLayer(this.xrSession);` or something. WebGPU backend simply returns null. Prevents us from having to shovel GL-specific data around.

src/main.ts
```diff
@@ -348,12 +370,24 @@ class Main {
...
         this.ui.update();
+    }
+
+    private _updateLoop = (time: number) => {
+        if (this.paused)
+            return;
+        
+        const updateInfo: ViewerUpdateInfo = {
+            time: time
+        }
```
magcius: Please don't create garbage in the hot path.

fordacious: ??

magcius: Creating new objects creates GC garbage.

src/ui.ts
```diff
@@ -2500,6 +2556,7 @@ export class UI {
     public sceneSelect: SceneSelect;
     public textureViewer: TextureViewer;
     public viewerSettings: ViewerSettings;
+    public xrSettings?: XRSettings;
```
magcius: `public xrSettings: XRSettings | null`. Don't use `undefined`.

src/ui.ts
```diff
@@ -1629,6 +1630,61 @@ class ViewerSettings extends Panel {
...
+        this.contents.appendChild(this.scaleSlider.elem);
+    }
+
+    private async enableXRChecked(saveManager: SaveManager, key: string) {
+        const enableXR = this.EnableXRCheckBox.checked;
+        this.EnableXRCheckBox.setChecked(enableXR);
+
+        if (this.EnableXRCheckBox.checked) {
+            try {
+                await this.webXRContext.start();
```
magcius: This should be driven by main. Provide callbacks like `onxrenabled()` that main connects to, rather than passing in the XR settings.

src/Camera.ts
```diff
@@ -437,6 +449,109 @@ export class FPSCameraController implements CameraController {
...
+            const cameraWorldMatrix = mat4.create();
+            cameraWorldMatrix.set(xrView.transform.matrix);
+            const cameraWorldMatrixTranslation = vec3.create();
+            mat4.getTranslation(cameraWorldMatrixTranslation, cameraWorldMatrix);
+            const originalViewTranslation = vec3.create();
+            mat4.getTranslation(originalViewTranslation, cameraWorldMatrix);
+
+            // Scale up view position and add offset
+            const cameraScale = vec3.create();
+            const cameraOrientation = quat.create();
```
magcius: Comments about GC garbage apply to the .create's in here as well. Use scratch vectors if absolutely necessary.

src/Camera.ts
```diff
@@ -437,6 +449,109 @@ export class FPSCameraController implements CameraController {
...
+            mat4.getScaling(cameraScale, cameraWorldMatrix);
+            mat4.getRotation(cameraOrientation, cameraWorldMatrix);
+
+            const cameraAdditionalOffset = vec3.create();
+            cameraAdditionalOffset.set(this.offset);
+            vec3.sub(cameraAdditionalOffset, cameraAdditionalOffset, originalViewTranslation);
+            vec3.scale(cameraWorldMatrixTranslation, cameraWorldMatrixTranslation, this.worldScale);
+            vec3.add(cameraWorldMatrixTranslation, cameraWorldMatrixTranslation, cameraAdditionalOffset);
```
magcius: `vec3.scaleAndAdd` might be a bit cleaner.

src/Camera.ts
```diff
@@ -437,6 +449,109 @@ export class FPSCameraController implements CameraController {
...
+            mat4.invert(camera.viewMatrix, camera.worldMatrix);
+            camera.worldMatrixUpdated();
+
+            // Unpack the projection matrix to get required parameters for setting clip / frustrum etc...
+            const cameraProjectionMatrix = xrView.projectionMatrix;
+            camera.projectionMatrix.set(cameraProjectionMatrix);
+            const fov = 2.0*Math.atan(1.0/cameraProjectionMatrix[5]);
+            const aspect = cameraProjectionMatrix[5] / cameraProjectionMatrix[0];
+            const shearX = cameraProjectionMatrix[8];
+            const shearY = cameraProjectionMatrix[9];
```
magcius: gross, but understandable.

src/Camera.ts
```diff
@@ -28,6 +30,13 @@ export class Camera {
     public orthoScaleY: number;
     public aspect: number;
     public isOrthographic: boolean = false;
+    public shearX: number = 0;
+    public shearY: number = 0;
+
+    private webXROverrideCameraProperties: boolean = false;
+    public setWebXROverrideEnabled(enabled: boolean): void {
+        this.webXROverrideCameraProperties = enabled;
+    }
```
magcius: I don't exactly like how this is done right now, but blaming it on you is wrong -- I think you can tell my camera code leaves a lot to be desired :)

fordacious: Yeah the strategy here was to be as minimal and non-disruptive as possible with the webxr changes and save digging deeply into the code for later

PR #244, opened 2020-02-28
Added a small help section in the VR Settings menu

MelonSpeedruns:


magcius:
This panel will have to be redesigned eventually. Fine with this for now.

PR #245, opened 2020-02-28
Added tutorial on how to enable WebXR on Chrome

MelonSpeedruns:


MelonSpeedruns:
Apparently you don't need to change settings in Chrome to get it working? I'm looking at that.

PR #260, opened 2020-03-17
Improve water and reflective surfaces in Star Fox Adventures

beholdnec:
This PR improves the look of water, and implements the reflective floors in Krazoa Palace (which looked rather bizarre in the game, too).

### DIFF
--- a/src/StarFoxAdventures/shaders.ts
+++ b/src/StarFoxAdventures/shaders.ts
@@ -241,12 +241,14 @@ function makeWaterRelatedTexture(device: GfxDevice): SFATexture {
     return { gfxTexture, gfxSampler, width, height }
 }
 
+type TexMtx = ((dst: mat4, viewerInput: ViewerRenderInput, modelMtx: mat4) => void) | undefined;
+
 export function buildMaterialFromShader(device: GfxDevice, shader: Shader, texColl: TextureCollection, texIds: number[], alwaysUseTex1: boolean, isMapBlock: boolean): SFAMaterial {
     const mb = new GXMaterialBuilder('Material');
     const textures = [] as SFAMaterialTexture[];
-    const texMtx: ((viewerInput: ViewerRenderInput, modelMtx: mat4) => mat4)[] = nArray(10, () => () => mat4.create());
-    const postTexMtx: (mat4 | null)[] = nArray(20, () => null);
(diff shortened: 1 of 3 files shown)


src/StarFoxAdventures/shaders.ts
```diff
@@ -638,6 +699,23 @@ export function buildMaterialFromShader(device: GfxDevice, shader: Shader, texCo
 
     return {
         material: mb.finish(),
-        textures
+        textures,
+        setupMaterialParams: (params: MaterialParams, viewerInput: ViewerRenderInput, modelViewMtx: mat4) => {
+            for (let i = 0; i < 10; i++) {
+                params.u_TexMtx[i] = texMtx[i](viewerInput, modelViewMtx);
```
magcius: Why not `texMtx[i](params.u_TexMtx[i], viewerInput, modelMatrix)` and prevent the extra copy?

src/StarFoxAdventures/shaders.ts
```diff
@@ -638,6 +699,23 @@ export function buildMaterialFromShader(device: GfxDevice, shader: Shader, texCo
...
+            }
+            
+            for (let i = 0; i < 3; i++) {
+                if (indTexMtx[i] !== null) {
+                    params.u_IndTexMtx[i] = indTexMtx[i]!;
```
magcius: I would prefer `mat4.copy` rather than re-assignment.

src/StarFoxAdventures/shaders.ts
```diff
@@ -560,13 +607,20 @@ export function buildMaterialFromShader(device: GfxDevice, shader: Shader, texCo
...
 
-        // TODO: GXSetIndTexMtx
+        indTexMtx[2] = mat4.create();
+        mat4.set(indTexMtx[2],
+            0.0, 0.5, 0.0, 0.0,
+            -0.5, 0.0, 0.0, 0.0,
+            0.0, 0.0, 1.0, 0.0,
+            0.0, 0.0, 0.0, 1.0
+        );
+        mat4.transpose(indTexMtx[2], indTexMtx[2]);
```
magcius: Could you just inline the transpose here? I think the only difference is where you put the negative sign.

beholdnec:
Updated to address feedback.

PR #325, opened 2020-06-28
Noclip Studio - Create Interpolated 3D Camera Animations

Veegie:
This PR introduces a new feature into noclip, one that I'm tentatively dubbing "Studio Mode".

This feature allows an end user to create re-playable interpolated 3D camera animations using the existing FPS Camera Controller controls in noclip, and a new control panel.

Apologies if this PR description is a bit wordy. If nothing else the text can be repurposed as documentation.

### How it works
The PR adds a new panel to the UI, denoted with a clapboard icon. From here the user can enable Studio Mode.

In Studio Mode, the user can move the camera and press Enter to place "Keyframes". A Keyframe simply describes a position and orientation of the camera, as well as some logic values for interpolating the camera to that state from wherever it currently is. So, an animation is an array of Keyframes. The newly-added StudioCameraController plays back an animation by iterating through this array and interpolating the camera to each position in turn, optionally looping back to the starting position at the end.

The use case for this feature presumes that the user will use external screen-capture software such as OBS to capture the actual video of their animation. There are a couple of quality of life options to facilitate this, like a delayed start, and automatically hiding the UI during playback.

### The Interface
The majority of the changes in this PR are UI changes. In spite of the large change-set, the UI is fairly utilitarian, as I tried to keep to the existing style and conventions of noclip. I was also told a UI overhaul for the entire site might be incoming sometime soon, so I didn't iterate very much on my design.

image
As this feature captures a rather niche use case, one of my main focuses was ensuring that as much of its functionality as possible remains uninitialized until the user elects to use it. Studio Mode is disabled by default, with its UI elements completely unloaded, not simply hidden.

---

image
Enabling Studio Mode loads in the UI elements, forces a change to the WASD controller if it's not already selected, and provides an "onboarding" prompt that teaches the user how to place a keyframe before showing the full interface.

---

image
Placing the start position keyframe reveals the full interface. The onboarding text is replaced with a default prompt.

---

image
Most elements have help text, which is displayed at the top of the panel when that element is moused over or in-focus. In the above image, the "Hold Duration" input is focused.

---

image
Adding another keyframe reveals some interpolation options. A quick rundown of all the options:

**Edit Position**: Allows the user to change the end position of the selected keyframe. Click to enter Edit mode, press Enter to save the new position.

**Name**: Just a QoL option to allow the user to name keyframes for logistical purposes. Purely front-end.

**Duration**: The time in seconds that the animation should take to interpolate to this keyframe's position from the last keyframe. Has decisecond precision (though it's worth noting the backend has millisecond precision).

**Interpolation**: Provides some bezier easing options for the interpolation of the camera motion, using CSS parlance.
 - Linear: Linear interpolation.
 - Ease In: Gradual acceleration at start, ending with linear speed.
 - Ease Out: Linear speed at start, ending with gradual deceleration.
 - Ease Both: Gradual acceleration and deceleration at start and end.

These interpolation buttons are hidden if Duration is set to 0 seconds, as there is no motion to interpolate.

**Hold Duration**: Time in seconds to hold at the end position of this keyframe before moving to the next keyframe. Same precision as Duration.

**Move Up/Down**: Re-arranges the selected keyframe in the keyframe list. Note that the Starting Position keyframe is immovable.

**Preview keyframe**: Plays a preview of the selected keyframe's current settings. Moves the camera to the end position of the previous keyframe, and plays back the interpolation. Useful for testing and fine-tuning motion in longer animations.

**Arrow Buttons**: Navigates through the keyframe list. Keyframes can also be selected by clicking on them manually.

Note that when a keyframe is selected, the camera jumps to that keyframe's end position. New keyframes are inserted *after the currently-selected keyframe*. Keyframes can be deleted by clicking the X button next to their name in the list.

---

image
And lastly, some options at the bottom for controlling playback of the full animation. Fairly self-explanatory. The one caveat is that enabling "Loop animation" also enables the interpolation options for the Starting Position keyframe. The animation will interpolate from the ending position of the last keyframe to the starting position using the starting position keyframe's settings.

### Example 
https://gfycat.com/incompleteclosebanteng - A facsimile of the intro camera movements from SM64, in SM64DS. A bit jank, especially between keyframes, but not too terrible considering this animation took me about 5 minutes to make.

### New dependency
This PR adds a small dependency, [bezier-easing](https://www.npmjs.com/package/bezier-easing). I notice that [Spline.ts](../../blob/master/src/Spline.ts) has a function `getPointBezier`, but bezier-easing seemed to be a bit more performant for this use case. I'm open to refactoring and using Spline.ts if necessary, of course.

### Possible Improvements
Numerous! I consider this to be a first-pass, [MVP](https://en.wikipedia.org/wiki/Minimum_viable_product)-type deal. If this feature is accepted and proves popular, there are a number of improvements I already have in mind, including:
 - More customizable interpolation (try to make an animation with 4 keyframes that rotates about the Y axis around an origin point and you'll see why this could be helpful),
 - Linking/grouping keyframes to produce consistent camera speed across multiple keyframes, 
 - Exporting/Importing animations to allow saving, loading and sharing without video recording,
 - UI Improvements.

Phew! That should about do it. Thanks for reading, and I hope y'all enjoy it. It's a pretty fun feature to fiddle around with.


magcius:
So first off, thank you! I know you've been working on this for a while. It will take me some time to review through all of it. And apologies in advance -- some of the comments might be a bit nitpicky.

The first thing I would like to suggest is that for now, I would prefer if this had to be enabled from the DevTools console. I promise that with the impending UI rewrite, we will have a bigger space for advanced usage and features like this, but for now, since UI is at a premium, I'm not sure it's a great fit to be enabled by default. Also, I don't want to place too much support workload on myself.

src/CameraAnimationManager.ts
```diff
@@ -0,0 +1,271 @@
+import { mat4 } from 'gl-matrix';
+import { Viewer } from './viewer';
+import { StudioCameraController } from './Camera';
+import BezierEasing from 'bezier-easing';
+
+const MAX_KEYFRAME_DURATION_SECONDS = 100.0;
+const MIN_KEYFRAME_DURATION = 0;
+const MILLISECONDS_IN_SECOND = 1000.0;
+
+export enum InterpolationType {
```
magcius, line 10: Nit: please use `const enum` where possible, and in newer code I've tried to adapt InterpolationType.Linear and so on as my preferred style. I also like using numbers rather than strings.

src/CameraAnimationManager.ts
```diff
@@ -0,0 +1,271 @@
...
+
+    /**
+     * A counter used for default keyframe names. Because keyframes can be deleted and
+     * inserted at arbitrary places, we can't simply rely on the current keyframe count 
+     * for this. The starting position keyframe is named differently, so we start at -1.
+     */
+    private _totalKeyframesAdded: number = -1;
+
+    constructor() {
+        this.keyframes = [];
```
magcius, line 141: You can remove this constructor by just using `keyframes: Keyframe[] = [];` in the definition above.

src/CameraAnimationManager.ts
```diff
@@ -0,0 +1,271 @@
...
+
+    appendKeyframe(keyframeEndPos: mat4) {
+        this.insertKeyframe(this.keyframes.length - 1, keyframeEndPos);
+    }
+
+    removeKeyframe(index: number) {
+        this.keyframes.splice(index, 1);
+    }
+
+    get totalKeyframesAdded(): number {
```
magcius, line 157: Personally not a huge fan of getters and setters. I would prefer to just leave this as a public property and access it from outside.

src/ui.ts
```diff
@@ -730,6 +732,7 @@ export class Panel implements Widget {
         this.toplevel.onkeydown = this.onKeyDown.bind(this);
         this.toplevel.onmouseover = this.syncSize.bind(this);
         this.toplevel.onmouseout = this.syncSize.bind(this);
+        this.toplevel.onclick = this.syncSize.bind(this);
```
magcius, line 735: Why is this necessary?

src/ui.ts
```diff
@@ -2677,7 +3280,7 @@ export class UI {
 
         this.isDragging = isDragging;
         this.elem.style.pointerEvents = (isDragging && !isPointerLocked) ? 'none' : '';
-        if (isDragging && this.shouldPanelsAutoClose())
+        if (isDragging && this.shouldPanelsAutoClose() && !this.studioModeEnabled)
```
magcius, line 3283: Move this condition into `shouldPanelsAutoClose()`

src/CameraAnimationManager.ts
```diff
@@ -0,0 +1,271 @@
...
+        } else {
+            this._holdDuration = newHoldDurationSeconds * MILLISECONDS_IN_SECOND;
+        }
+    }
+
+    get interpType(): InterpolationType {
+        return this._interpType;
+    }
+
+    set interpType(type: InterpolationType) {
```
magcius, line 97: Have you considered using something like a Hermite spline? That way you can have a smooth path that won't break, and only needs a couple extra parameters at each keyframe (tangent in/tangent out, but they can be locked to the same value for a smooth curve).

I think that would get much better results than treating each curve segment as an independent ease.

src/ui.ts
Veegie: I put an explanation in the commit message. It fixes a couple quirks with the Panels updating their sizes. For example, when changing the Camera Controller, the panel resizes immediately when showing/hiding options (e.g., switching to WASD from either of the other two), rather than having to mouse off the camera controller button first.

I initially added this because without it, when the studio panel was first initialized, it wouldn't expand to show all of its contents until it was moused over. However, now that I'm using a FloatingPanel, this is no longer an issue regardless. I can remove this if you'd like.

src/CameraAnimationManager.ts
Veegie: I was initially using numbers for this, but found that switching to strings made the [button logic](https://github.com/magcius/noclip.website/blob/9dac2d4b33dda6f38f1564e25ba547492505af82/src/ui.ts#L2194) cleaner. It's not hard to just parse an Int instead, though, so I'm fine to change back to numbers if you'd prefer.

src/CameraAnimationManager.ts
Veegie: I'll be honest -- I've never really worked with 3D graphics outside of a single introductory university course five years ago. I had no idea what a Hermite spline was until you mentioned it here! I've looked into it and it seems like exactly what I was planning to do with respect to "grouping keyframes" under Possible Improvements.

I'll keep looking into it with the intent of adding it to this PR, as I imagine that'll help to clear up the jank in the animations I've made so far.

Veegie:
Addressed the initial review comments, including hiding the Studio Panel by default. It can now be shown with a console call to `main.ui.studioPanel.s()`

Thanks for the tip to turn it into a `FloatingPanel`, as well -- there's a lot more room to make use of now and in the future, and it's definitely more user-friendly, to boot.

Veegie:
Phew. That should do for now. I do have my doubts about my method of tangent calculation (to say nothing of that code's readability, which is... oof), though it produces pretty nice results for me unless the keyframes are very awkwardly-placed. That's one reason I kept linear interpolation as a secondary option; sometimes all you want is a straightforward dolly, after all.

package.json
```diff
@@ -35,6 +35,7 @@
   "dependencies": {
     "@sentry/browser": "^5.7.1",
     "@types/webmidi": "^2.0.4",
+    "bezier-easing": "^2.1.0",
```
magcius, line 38: I'd prefer not to add any additional dependencies, if possible. Would it be possible to use Hermite for time interpolation as well?

src/CameraAnimationManager.ts
```diff
@@ -0,0 +1,371 @@
...
+            this._holdDuration = newHoldDurationSeconds * MILLISECONDS_IN_SECOND;
+        }
+    }
+
+    get linearEaseType(): LinearEaseType {
+        return this._linearEaseType;
+    }
+
+    set linearEaseType(type: LinearEaseType) {
+        switch (type) {
```
magcius, line 119: I prefer if statement chains to switch statements these days.

src/CameraAnimationManager.ts
```diff
@@ -0,0 +1,371 @@
...
+    }
+
+}
+
+export class CameraAnimation {
+    /**
+     * The list of keyframes that comprise this animation. Note that the `endPosition` 
+     * of the keyframe at index 0 is the camera's starting position for the animation.
+     */
+    keyframes: Keyframe[] = [];
```
magcius, line 152: private

src/CameraAnimationManager.ts
```diff
@@ -0,0 +1,371 @@
...
+
+    /**
+     * A counter used for default keyframe names. Because keyframes can be deleted and
+     * inserted at arbitrary places, we can't simply rely on the current keyframe count 
+     * for this. The starting position keyframe is named differently, so we start at -1.
+     */
+    public totalKeyframesAdded: number = -1;
+
+    insertKeyframe(after: number, keyframeEndPos: mat4) {
```
magcius, line 161: public

src/CameraAnimationManager.ts
```diff
@@ -0,0 +1,371 @@
...
+
+    public usesLinearInterp: boolean = false;
+    private _linearEaseType: LinearEaseType = LinearEaseType.EaseBoth;
+    private easingFunction: BezierEasing.EasingFunction | null = easeBothFunc;
+    public trsTangentIn: vec3;
+    public trsTangentOut: vec3;
+
+    constructor(public endPos: mat4) {
+        this.trsTangentIn = vec3.create();
+        this.trsTangentOut = vec3.create();
```
magcius, line 43: You can just do `public trsTangentOut = vec3.create();` above

src/CameraAnimationManager.ts
```diff
@@ -0,0 +1,371 @@
...
+    }
+
+    /**
+     * Updates the current state of this Keyframe.
+     * 
+     * @param dt delta time since last update
+     */
+    public update(dt: number): void {
+        if (this._interpProgress < this._interpDuration) {
```
magcius, line 52: It would probably be easier to keep a single time value.

src/CameraAnimationManager.ts
```diff
@@ -0,0 +1,371 @@
...
+        let startPos: mat4;
+        if (index > 0) {
+            startPos = this.currentAnimation.keyframes[index - 1].endPos;
+        } else if (index === 0) {
+            startPos = this.currentAnimation.keyframes[this.currentAnimation.keyframes.length - 1].endPos;
+        } else {
+            return;
+        }
+        this.calculateTangents();
+        this.studioCameraController.playAnimation(new Array(this.currentAnimation.keyframes[index]), startPos, false);
```
magcius, line 272: `playAnimation([this.currentAnimation.keyframes[index]), ...)`

src/CameraAnimationManager.ts
```diff
@@ -0,0 +1,371 @@
...
+        }
+    }
+
+    public previewKeyframe() {
+        const index = this.selectedKeyframeIndex;
+        let startPos: mat4;
+        if (index > 0) {
+            startPos = this.currentAnimation.keyframes[index - 1].endPos;
+        } else if (index === 0) {
+            startPos = this.currentAnimation.keyframes[this.currentAnimation.keyframes.length - 1].endPos;
```
magcius, line 267: Don't we only want this if it loops? I don't get the motivation here.

src/CameraAnimationManager.ts
```diff
@@ -0,0 +1,371 @@
...
+
+    get interpAmount(): number {
+        if (this.usesLinearInterp && this.easingFunction)
+            return this.easingFunction(this._interpProgress / this._interpDuration);
+        else
+            return this._interpProgress / this._interpDuration;
+    }
+
+    get durationInSeconds(): number {
+        return this._interpDuration / MILLISECONDS_IN_SECOND;
```
magcius, line 87: Can you just store these fields in seconds?

src/CameraAnimationManager.ts
```diff
@@ -0,0 +1,371 @@
...
+     * The length of time in milliseconds it should take to animate to this Keyframe's end position.
+     */
+    private _interpDuration: number = 5000.0;
+    /**
+     * The length of time in milliseconds to hold on this keyframe's end position before moving to the next keyframe.
+     */
+    private _holdDuration: number = 0.0;
+
+    private _interpProgress: number = 0;
+    private _holdProgress: number = 0;
```
magcius, line 33: I'm not a huge fan of how the Keyframe objects record state. I would prefer for them to just be data objects, and any state is kept on the CameraAnimationManager. That would make it easier to import/export animations.

src/Camera.ts
```diff
@@ -496,6 +498,151 @@ export class FPSCameraController implements CameraController {
...
+    public playAnimation(keyframes: Keyframe[], startPos: mat4, loop: boolean) {
+        this.animationKeyframes = keyframes;
+        this.isAnimationPlaying = true;
+        this.loopAnimation = loop;
+        this.currentKeyframeIndex = -1;
+        this.setToPosition(startPos);
+
+        // If the starting position is the same as the first keyframe's
+        // end position, skip any interpolation for that keyframe.
+        if (mat4.exactEquals(startPos, keyframes[0].endPos)) {
```
magcius, line 612: When would this not be the case? When previewing keyframes?

src/Camera.ts
```diff
@@ -496,6 +498,151 @@ export class FPSCameraController implements CameraController {
...
+            result = super.update(inputManager, dt);
+            if (this.isOnKeyframe && result !== CameraUpdateResult.Unchanged) {
+                this.isOnKeyframe = false;
+            }
+        }
+
+        return result;
+    }
+
+    updateAnimation(dt: number): CameraUpdateResult {
```
magcius, line 555: public

src/CameraAnimationManager.ts
Veegie: Yes. You can only preview the animation for keyframe 0 if looping is enabled. Previewing Keyframe 0 shows you the motion from the last keyframe back to the start position.

Veegie:
There we go. Moving all of the animation and keyframe state logic to CameraAnimationManager definitely makes a lot more sense.

Also, I didn't say this before, but thanks for taking the time to review this PR!

src/CameraAnimationManager.ts
```diff
@@ -0,0 +1,415 @@
...
+
+    constructor(public endPos: mat4) {
+    }
+
+    get interpDuration(): number {
+        return this._interpDuration;
+    }
+
+    set interpDuration(d: number) {
+        if (d <= MIN_KEYFRAME_DURATION)
```
magcius, line 52: If we want this to be a pure data object, I'd prefer to remove the getters/setters -- I feel like this is more "UI" than core data structures. Also, you can just use `clamp(d, MIN_KEYFRAME_DURATION, MAX_KEYFRAME_DURATION)`

src/CameraAnimationManager.ts
```diff
@@ -0,0 +1,415 @@
...
+    set linearEaseType(type: LinearEaseType) {
+        if (type === LinearEaseType.NoEase)
+            this.easingFunction = null;
+        else if (type === LinearEaseType.EaseIn)
+            this.easingFunction = easeInFunc;
+        else if (type === LinearEaseType.EaseOut)
+            this.easingFunction = easeOutFunc;
+        else if (type === LinearEaseType.EaseBoth)
+            this.easingFunction = easeBothFunc;
+        this._linearEaseType = type;
```
magcius, line 86: For similar reasons, mind moving the easing function elsewhere? Shouldn't be difficult to add a `getEasingFunctionForEaseType` helper and call that from CAM. Bonus points if you make this an interface so it can JSON.stringify easier.

Veegie:
Done and done!

magcius:
Looks good to me!

PR #333, opened 2020-08-05
Studio - Interpolation Improvements

Veegie:
This PR introduces some improvements to the interpolation methods used in Studio mode.

- A bug in the playback engine has been fixed. Now, if the delta time between one animation step and the next overruns the current keyframe's duration, we save that value and carry it over to the next keyframe. I've done some testing and haven't noticed any issues with this yet, and it should help to reduce judder on keyframe boundaries.
- Interpolation is now done via the generation of a lookAt matrix (or targetTo matrix, more precisely) every animation step, so that we can use Hermite interpolation on both rotation and translation. The difference between [quaternion slerp](https://gfycat.com/neighboringultimateblacknorwegianelkhound) and [Hermite](https://gfycat.com/charmingaromaticivorybackedwoodswallow) is subtle, but the latter is definitely an improvement. Big thanks to Jasper for all their help on this one.
- A new button has been added to the interface, which, when clicked, will attempt to calculate the duration necessary to match the preceding keyframe's interpolation speed.
   - It estimates the curve length of a segment between two keyframes by taking 10,000 sample points along the curve, calculates the euclidean distance between each point, and sums them all together.
   - The duration of the previous keyframe is divided by its curve length, then multiplied by the current keyframe's curve length to retrieve an approximate duration to match the (average) speed. This doesn't account for acceleration, just a constant average speed, but it's a decent starting point for a user to smooth things out manually.
- Fixed a bug where the starting position couldn't be edited.

**Limitations**
- Keyframes consisting only of rotation are a bit wonky with Hermite interpolation owing to the unwieldy tangents. This might be fixable with configurable focus distances, but I'm not sure. Linear interpolation is a good alternative for these kinds of keyframes.
- (Fixed in 78cbb7f) ~~The lookAt/targetTo interpolation method does not support bank rotation, ("roll rotation" about the z-axis) because math is really hard. If anyone knows how to make math less hard, please email me at my webzone.~~
- (Fixed in 10aa85f) ~~Animations saved with the previous system are incompatible with this update. Animations saved in localStorage will be deleted upon enabling studio mode, but the JSON will be printed to the console before deleting. If you've exported an animation to JSON using the previous interpolation method, you'll need to do two search-and-replace edits to make them compatible with this new system:~~
   - ~~Find `trsTangent`, replace all with `posTangent`~~
   - ~~Find `rotTangent`, replace all with `lookAtPosTangent`~~

### DIFF
--- a/src/CameraAnimationManager.ts
+++ b/src/CameraAnimationManager.ts
@@ -209,7 +209,7 @@ export class CameraAnimationManager {
     }
 
     public editKeyframePosition(pos: mat4, index: number) {
-        if (index > -1 && index < this.animation.keyframes.length)
+        if (index >= 0 && index < this.animation.keyframes.length)
             this.animation.keyframes[index].endPos = pos;
         this.endEditKeyframePosition();
     }
--- a/src/CameraAnimationManager.ts
+++ b/src/CameraAnimationManager.ts
@@ -404,9 +404,10 @@ export class CameraAnimationManager {
         let length = 0;
         if (!vec3.exactEquals(this.prevPos, this.nextPos)) {
             vec3.copy(this.scratchVec1, this.prevPos);
-            for (let t = 0.0001; t <= 1; t += 0.0001) {
+            const numSteps = 10000;
+            for (let i = 1; i <= numSteps; i++) {
                 for (let j = 0; j < 3; j++) {
-                    this.scratchVec2[j] = getPointHermite(this.prevPos[j], this.nextPos[j], toKeyframe.posTangentIn[j], toKeyframe.posTangentOut[j], t);
+                    this.scratchVec2[j] = getPointHermite(this.prevPos[j], this.nextPos[j], toKeyframe.posTangentIn[j], toKeyframe.posTangentOut[j], i / numSteps);
                 }
                 length += vec3.distance(this.scratchVec1, this.scratchVec2);
                 vec3.copy(this.scratchVec1, this.scratchVec2);
--- a/src/CameraAnimationManager.ts
+++ b/src/CameraAnimationManager.ts
@@ -591,30 +591,41 @@ export class CameraAnimationManager {
             return;
         }
 
+        // Speed scales calculated as per Nils Pipenbrinck:
+        // https://www.cubic.org/docs/hermite.htm - section "Speed Control".
+        const outScale =  (2 * curKf.interpDuration) / (curKf.interpDuration + nextKf.interpDuration);
+        const inScale = (2 * nextKf.interpDuration) / (curKf.interpDuration + nextKf.interpDuration);
+
         vec3.set(this.prevPos, prevKf.targetPositionX.value, prevKf.targetPositionY.value, prevKf.targetPositionZ.value);
         vec3.set(this.nextPos, nextKf.targetPositionX.value, nextKf.targetPositionY.value, nextKf.targetPositionZ.value);
         vec3.sub(this.scratchVec1, this.nextPos, this.prevPos);
         vec3.scale(this.scratchVec1, this.scratchVec1, this.scaleFactor);
+        vec3.copy(this.scratchVec2, this.scratchVec1);
+        vec3.scale(this.scratchVec1, this.scratchVec1, outScale);
+        vec3.scale(this.scratchVec2, this.scratchVec2, inScale);
         curKf.targetPositionX.tangentOut = this.scratchVec1[0];
         curKf.targetPositionY.tangentOut = this.scratchVec1[1];
         curKf.targetPositionZ.tangentOut = this.scratchVec1[2];
-        nextKf.targetPositionX.tangentIn = this.scratchVec1[0];
-        nextKf.targetPositionY.tangentIn = this.scratchVec1[1];
-        nextKf.targetPositionZ.tangentIn = this.scratchVec1[2];
+        nextKf.targetPositionX.tangentIn = this.scratchVec2[0];
+        nextKf.targetPositionY.tangentIn = this.scratchVec2[1];
+        nextKf.targetPositionZ.tangentIn = this.scratchVec2[2];
 
         vec3.set(this.prevLookAtPos, prevKf.lookAtPositionX.value, prevKf.lookAtPositionY.value, prevKf.lookAtPositionZ.value);
         vec3.set(this.nextLookAtPos, nextKf.lookAtPositionX.value, nextKf.lookAtPositionY.value, nextKf.lookAtPositionZ.value);
-        vec3.sub(this.scratchVec2, this.nextLookAtPos, this.prevLookAtPos);
-        vec3.scale(this.scratchVec2, this.scratchVec2, this.scaleFactor);
-        curKf.lookAtPositionX.tangentOut = this.scratchVec2[0];
-        curKf.lookAtPositionY.tangentOut = this.scratchVec2[1];
-        curKf.lookAtPositionZ.tangentOut = this.scratchVec2[2];
+        vec3.sub(this.scratchVec1, this.nextLookAtPos, this.prevLookAtPos);
+        vec3.scale(this.scratchVec1, this.scratchVec1, this.scaleFactor);
+        vec3.copy(this.scratchVec2, this.scratchVec1);
+        vec3.scale(this.scratchVec1, this.scratchVec1, outScale);
+        vec3.scale(this.scratchVec2, this.scratchVec2, inScale);
+        curKf.lookAtPositionX.tangentOut = this.scratchVec1[0];
+        curKf.lookAtPositionY.tangentOut = this.scratchVec1[1];
+        curKf.lookAtPositionZ.tangentOut = this.scratchVec1[2];
         nextKf.lookAtPositionX.tangentIn = this.scratchVec2[0];
         nextKf.lookAtPositionY.tangentIn = this.scratchVec2[1];
         nextKf.lookAtPositionZ.tangentIn = this.scratchVec2[2];
 
-        curKf.bank.tangentOut = (nextKf.relativeBank - prevKf.relativeBank) * this.scaleFactor;
-        nextKf.bank.tangentIn = (nextKf.relativeBank - prevKf.relativeBank) * this.scaleFactor;
+        curKf.bank.tangentOut = (nextKf.relativeBank - prevKf.relativeBank) * this.scaleFactor * outScale;
+        nextKf.bank.tangentIn = (nextKf.relativeBank - prevKf.relativeBank) * this.scaleFactor * inScale;
     }
 
     private zeroEndpointTangents(): void {
--- a/src/CameraAnimationManager.ts
+++ b/src/CameraAnimationManager.ts
@@ -143,6 +143,12 @@ export class CameraAnimationManager {
     public loadAnimation(keyframes: Keyframe[]) {
         if (keyframes.length === 0)
             return;
+        for (let i = 0; i < keyframes.length; i++) {
+            keyframes[i].posTangentIn = vec3.create();
+            keyframes[i].posTangentOut = vec3.create();
+            keyframes[i].lookAtPosTangentIn = vec3.create();
+            keyframes[i].lookAtPosTangentOut = vec3.create();
+        }
         this.animation = new CameraAnimation();
         this.animation.keyframes = keyframes;
         this.uiKeyframeList.dispatchEvent(new Event('startPositionSet'));
@@ -159,7 +165,13 @@ export class CameraAnimationManager {
     }
 
     public serializeAnimation(): string {
-        return JSON.stringify(this.animation.keyframes);
+        const exclude = ['posTangentIn', 'posTangentOut', 'lookAtPosTangentIn', 'lookAtPosTangentOut'];
+        return JSON.stringify(this.animation.keyframes, (key, value) => {
+            if (exclude.includes(key))
+                return undefined;
+            else
+                return value;
+        });
     }
 
     public getKeyframeByIndex(index: number): Keyframe {
@@ -402,11 +414,6 @@ export class CameraAnimationManager {
                 for (let j = 0; j < 16; j++) {
                     if (typeof a[i].endPos[j] !== 'number')
                         return false;
-                    if (j < 3) {
-                        if (typeof a[i].posTangentIn[j] !== 'number'
-                            || typeof a[i].posTangentOut[j] !== 'number')
-                            return false;
-                    }
                 }
                 if (typeof a[i].holdDuration !== 'number')
                     return false;
--- a/src/CameraAnimationManager.ts
+++ b/src/CameraAnimationManager.ts
@@ -87,6 +87,7 @@ export class CameraAnimationManager {
      * The translation vector components of the keyframes following and preceding the current keyframe.
      * Used for calculating tangents.
      */
+    private beforePrevTrs: vec3 = vec3.create();
     private prevTrs: vec3 = vec3.create();
     private nextTrs: vec3 = vec3.create();
     private tangentScratchVec: vec3 = vec3.create();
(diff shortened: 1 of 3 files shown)


Veegie:
Sorry, I've heard it's bad form to push extra commits after opening a PR, but I realized that my speed estimation method was naive to the point of being completely useless, so I refined it. Now it's slightly less naive and a lot more accurate. Still not perfect, but it's actually useful now.

Also fixed an issue with tangent calculation for looping vs non-looping animations. And fixed the compatibility issue by just not bothering to stringify tangent values.

magcius:
Sorry for the delay on reviewing this -- so many projects to juggle right now. I'll try to take a look tonight.

Veegie:
> Sorry for the delay on reviewing this -- so many projects to juggle right now. I'll try to take a look tonight.

No worries, and no rush. The delay spared you from having to see my initial speed-matching approach, heh.

> I feel like they shouldn't be part of the Keyframe object in general if they're going to be recomputed on playback. It would be nice to have tangent values editable by the user.

I figured I would eventually make tangent values editable by the user, which is one reason why I didn't remove them from the Keyframe interface, since we'd just be adding them back later on anyway so the custom tangent values could be saved and loaded.

That, and I feel it's a little cleaner to keep the values on the keyframes and just read them during playback rather than having to store them elsewhere and go fetch them.

Veegie:
Bank rotation is now supported! Many thanks to @ageo-math for their invaluable assistance with both finding the correct bank rotation value and simplifying the logic for interpolation between bank values.

Also I finally cleaned up the tangent calculation logic. And with that, I promise I won't mess with this again until it's reviewed, heh.

src/CameraAnimationManager.ts
```diff
@@ -23,6 +24,12 @@ const easeBothFunc: Function = (t: number) => {
     return getPointBezier(0, 0, 1, 1, t);
 }
 
+export interface InterpolationStep {
```
magcius, line 27: Making this a class might be a tiny bit cleaner.

src/CameraAnimationManager.ts
```diff
@@ -34,9 +41,14 @@ export interface Keyframe {
     holdDuration: number;
     usesLinearInterp: boolean;
     linearEaseType: LinearEaseType;
```
magcius, line 43: Do we still need these?

src/CameraAnimationManager.ts
```diff
@@ -34,9 +41,14 @@ export interface Keyframe {
...
-    trsTangentIn: vec3;
-    trsTangentOut: vec3;
+    posTangentIn: vec3;
+    posTangentOut: vec3;
+    lookAtPosTangentIn: vec3;
+    lookAtPosTangentOut: vec3;
+    bankTangentIn: number;
+    bankTangentOut: number;
     endPos: mat4;
```
magcius, line 50: We can bikeshed a bit over the keyframe data format, but a cleanup might be something like:

```ts
{
   targetPosition: { value: ..., tangentIn: ..., tangentOut: ..., },
   lookAtPosition: { value: ..., tangentIn: ..., tangentOut: ..., },
   bank: { value: ..., tangentIn: ..., tangentOut: ..., },
}
```

That's relatively easy to support, and the data model a lot clearer, IMO. Then we don't need to decompose from endPos.

src/CameraAnimationManager.ts
```diff
@@ -34,9 +41,14 @@ export interface Keyframe {
     holdDuration: number;
     usesLinearInterp: boolean;
     linearEaseType: LinearEaseType;
-    trsTangentIn: vec3;
-    trsTangentOut: vec3;
+    posTangentIn: vec3;
+    posTangentOut: vec3;
+    lookAtPosTangentIn: vec3;
+    lookAtPosTangentOut: vec3;
```
magcius, line 47: Probably worth watching out for -- vec3 will serialize to JSON as `{"0": 0, "1": 0, "2": 0}`, and unserializing it will return an object, not a Float32Array. I think in this case everything will work out, because gl-matrix only uses the numbered properties and we don't use any of the specific array methods here, but it might be slower after a serialize/deserialize (since it will be backed by an Object, not a generic Array). Having a serialize/deserialize that ensures that these things are vec3's might be a good idea.

src/CameraAnimationManager.ts
```diff
@@ -135,7 +171,13 @@ export class CameraAnimationManager {
     }
 
     public serializeAnimation(): string {
-        return JSON.stringify(this.animation.keyframes);
+        const exclude = ['posTangentIn', 'posTangentOut', 'lookAtPosTangentIn', 'lookAtPosTangentOut, bankTangentIn, bankTangentOut, bank'];
```
magcius, line 174: You're excluding a key called `'lookAtPosTangentOut, bankTangentIn, bankTangentOut, bank'`, here :)

I don't know how much we care about stability in the input/output format, but at least I'd start to wrap it in a ```js
{ version: 1, keyframes: [...] }` object for now, so we don't have an ad-hoc data model going forward, and the legacy version is considered "version 0" :). We might not need it, but it doesn't hurt to give ourselves the option to have a deserializeVersion0, deserializeVersion1 in the future if we change the data model considerably.

src/CameraAnimationManager.ts
```diff
@@ -306,6 +361,60 @@ export class CameraAnimationManager {
...
+     * @param fromKeyframe the keyframe describing the start position
+     * @param toKeyframe the keyframe describing the end position
+     */
+    private estimateHermiteCurveLength(fromKeyframe: Keyframe, toKeyframe: Keyframe): number {
+        mat4.getTranslation(this.prevPos, fromKeyframe.endPos);
+        mat4.getTranslation(this.nextPos, toKeyframe.endPos);
+        let length = 0;
+        if (!vec3.exactEquals(this.prevPos, this.nextPos)) {
+            vec3.copy(this.scratchVec1, this.prevPos);
+            for (let t = 0.0001; t <= 1; t += 0.0001) {
```
magcius, line 407: I'd probably write this as:

```ts
const numSteps = 10000;
for (let i = 1; i < numSteps; i++) {
    const t0 = (i - 1) / numSteps, t1 = i / numSteps;
    getPointHermite3(scratchVec3a, fromKeyframe.prevPos, toKeyframe.nextPos, toKeyframe.posTangentIn, toKeyframe.posTangentOut, t0);
    getPointHermite3(scratchVec3b, fromKeyframe.prevPos, toKeyframe.nextPos, toKeyframe.posTangentIn, toKeyframe.posTangentOut, t1);
    length += vec3.distance(scratchVec3a, scratchVec3b);
}
```

Also, Super Mario Galaxy uses a numSteps of 10, so you might be overdoing it a bit with 10k :), but ALU is cheap anyway.

src/CameraAnimationManager.ts
```diff
@@ -348,81 +452,127 @@ export class CameraAnimationManager {
...
+
+    private calculateBankRotationValues() {
+        let previousBank = 0;
+        let fullRotations = 0;
+        let kf: Keyframe;
+        for (let i = 0; i < this.animation.keyframes.length; i++) {
+            kf = this.animation.keyframes[i];
+            computeEulerAngleRotationFromSRTMatrix(this.scratchVec1, kf.endPos);
+            vec3.copy(this.scratchVec2, Vec3UnitY);
+            vec3.rotateZ(this.scratchVec2, this.scratchVec2, Vec3Zero, -this.scratchVec1[2]);
```
magcius, line 480: I'm still a bit shocked this isn't the same as computeModelMatrixR, but I'd have to play around with it for a bit, I guess, and look at that robotics paper a bit more carefully. Good to know this works.

src/ui.ts
```diff
@@ -2029,6 +2033,8 @@ class StudioPanel extends FloatingPanel {
         this.keyframeNameInput = this.contents.querySelector('#keyframeName') as HTMLInputElement;
         this.keyframeDurationContainer = this.contents.querySelector('#keyframeDurationContainer') as HTMLElement;
         this.keyframeDurationInput = this.contents.querySelector('#keyframeDuration') as HTMLInputElement;
+        this.matchPrevSpeedBtn = this.contents.querySelector('#matchPrevSpeedBtn') as HTMLInputElement;
+        this.matchPrevSpeedBtn.dataset.helpText = 'Attempt to auto-calculate the duration to match the previous keyframe\'s speed. Not guaranteed to work.';
```
magcius, line 2037: Why isn't it guaranteed to work?

src/CameraAnimationManager.ts
magcius: Other options if you prefer some of these. Anybody up for a game of darts?

```ts
{
    targetPositionX: { value: ..., tangentIn: ..., tangentOut: ..., },
    targetPositionY: { value: ..., tangentIn: ..., tangentOut: ..., },
    targetPositionZ: { value: ..., tangentIn: ..., tangentOut: ..., },
    ...
}
```

```ts
{
    targetPositionX: {[
        { time: 0, value: ..., tangentIn: ..., tangentOut: ..., },
        { time: 30, value: ..., tangentIn: ..., tangentOut: ..., },
        { time: 60, value: ..., tangentIn: ..., tangentOut: ..., },
    ]},
    targetPositionY: {[
        { time: 0, value: ..., tangentIn: ..., tangentOut: ..., },
        { time: 30, value: ..., tangentIn: ..., tangentOut: ..., },
        { time: 60, value: ..., tangentIn: ..., tangentOut: ..., },
    ]},
    ...
}
```

The former "collapses" the vec3 down so that each X/Y/Z is now their own set of values and tangents, meaning it's all 1D and there's no more vec3's in the data model.

The latter uses a series of "animation tracks" with their own set of keyframes, which matches what traditional animation software like Blender, Maya and After Effects do (you can even have separate timing for X/Y/Z!). The benefit of this latter approach is that we might even be able to interpret camera animation data from some games. e.g. check out the Wii U Transfer Tool and https://github.com/magcius/noclip.website/blob/4b117ca3c7007619f2b29ec8d89f1ae0d939bf6e/src/rres/brres.ts#L3185-L3220 . Nintendo games tend to use a data model like this.

src/CameraAnimationManager.ts
```diff
@@ -167,7 +209,7 @@ export class CameraAnimationManager {
     }
 
     public editKeyframePosition(pos: mat4, index: number) {
-        if (index > 0 && index < this.animation.keyframes.length)
+        if (index > -1 && index < this.animation.keyframes.length)
```
magcius, line 212: Super minor nit but I tend to prefer `index >= 0`. Though I might recommend an assert here -- how are you getting out-of-bounds values to begin with?

src/CameraAnimationManager.ts
```diff
@@ -306,6 +361,60 @@ export class CameraAnimationManager {
...
+            const prevLength = this.estimateHermiteCurveLength(beforePrevKf, prevKf);
+            if (prevLength === 0)
+                return 0;
+            const ratio = prevKf.interpDuration / prevLength;
+            const curLength = this.estimateHermiteCurveLength(prevKf, curKf);
+            if (curLength > 0)
+                duration = ratio * curLength;
+        }
+        return duration;
+    }
```
magcius, line 393: I wonder if something people might want is a "match duration across curve segments". That is, I want these three keyframes to take up, say, 2 seconds of time, with the time split naturally according to the curve length.

src/ui.ts
Veegie: Because I forgot to update the help text, mostly. 🤔

But also I've had some mixed results even with the more accurate curve length estimation. I'm still looking into it.

Veegie:
Settled on the second, "collapsed" version of the three datamodel updates you suggested. The animation tracks felt like a bit too much at the moment, but I'll look into it for a future update. Shouldn't be a huge leap from the collapsed version to the tracks. Not as huge as this current leap, anyway.

Also, this isn't quite ready for review again just yet, I don't think. I'm still noticing some weird issues with tangent recalculation. Mostly just pushing these commits now so I can work from my other computer.

src/CameraAnimationManager.ts
Veegie: That was something I was planning to add, probably for the next big chunk of work on this. Mostly I need to figure out how to express that in the UI.

src/CameraAnimationManager.ts
Veegie: Chalk it up to excessive safety-checking learnt from half a decade of corporate software dev, I guess. Plus when no keyframe is selected, index is `-1`. Though ofc, if no keyframe is selected, Edit Position isn't clickable anyway...

src/CameraAnimationManager.ts
```diff
@@ -41,14 +41,13 @@ export interface Keyframe {
...
-    bankTangentOut: number;
-    endPos: mat4;
-    bank: number;
+    targetPositionX: { value: number, tangentIn: number, tangentOut: number };
+    targetPositionY: { value: number, tangentIn: number, tangentOut: number };
+    targetPositionZ: { value: number, tangentIn: number, tangentOut: number };
+    lookAtPositionX: { value: number, tangentIn: number, tangentOut: number };
+    lookAtPositionY: { value: number, tangentIn: number, tangentOut: number };
+    lookAtPositionZ: { value: number, tangentIn: number, tangentOut: number };
+    bank: { value: number, tangentIn: number, tangentOut: number };
```
magcius, line 50: You should be able to factor this out as either

```ts
type KeyframeTrack = { value: number, tangentIn: number, tangentOut: number };
```

or

```ts
interface KeyframeTrack { value: number, tangentIn: number, tangentOut: number }
```

(the two forms are equivalent, AFAIK)

Veegie:
Made a few more changes as per our discussion. I found a promising method to scale the auto-calculated tangents based on keyframe duration to make the speed changes less jarring, and this seems to fix my "Match previous speed" button issues as well. I've done a bit of testing and it seems to work fairly well in all of the animations I'd saved. Leave it to the late-90s demoscene to have better info on 3D graphics than most Wikipedia articles.

I also noted that looping animations had trouble with bank rotations when one or more full rotations is performed in the course of the animation (as it would interpolate from 360+ degrees back to ~0 or whatever the starting bank value was). The best way to fix this that I could figure was to just detect this edge case and modulo out however many full rotations had been performed up to the last keyframe. Then just interpolate back to the start from that modulo'd value instead. Seems to work.

Luckily, this bug also led to me discovering another issue, which was that my calculation of relative bank rotation values was modifying the *actual* bank rotation values on the keyframes. So, now I keep a separate value for the relative bank rotation on the keyframe.

src/CameraAnimationManager.ts
Veegie: Sorry, I missed this comment on the last pass-through.

These options allow the user to use a plain old linear interpolation instead of the Hermite curve on a per-keyframe basis. Basically equivalent to setting the tangent values to 0 for that keyframe, but `linearEaseType` gives a little fine-tuning on the t-value.

I figure, it's an option. It'll (probably?) be rendered obsolete when fine-tuning of tangent values and sub-keyframe timing becomes possible. But worth keeping for now.

Also, it remains the best way to interpolate a rotation-only keyframe for some reason.

magcius:
> I also noted that looping animations had trouble with bank rotations when one or more full rotations is performed in the course of the animation (as it would interpolate from 360+ degrees back to ~0 or whatever the starting bank value was).

Interesting... have you been keeping unnormalized rotation values as a feature, or by accident? Could we just normalize the rotations instead? Keeping the runtime model as close to pure Hermite and moving the smarts into the editor is where I'd prefer this to go.

> So, now I keep a separate value for the relative bank rotation on the keyframe.

I'm not the biggest fan of storing any runtime data on the Keyframe itself. The ideal is if CameraAnimationManager runtime uses a `Readonly<Keyframe>`, and all modification of it is done solely by the UI.

magcius:
Going to merge this for now, since it's a big improvement over what's there already. Feel free to make suggested improvements when you get around to it!

PR #337, opened 2020-08-29
Fix some inaccuracies in N64 color combiner implementation

RolandMunsil:
Also left in some helper methods I wrote for color combiner debugging

(If there's a better spot to put those helper methods I can move them ofc, just wasn't sure where the best place for them was)

### DIFF
--- a/src/BanjoKazooie/render.ts
+++ b/src/BanjoKazooie/render.ts
@@ -62,6 +62,7 @@ varying vec4 v_Color;
 varying vec4 v_TexCoord;
 
 const vec4 t_Zero = vec4(0.0);
+const vec4 t_Half = vec4(0.5);
 const vec4 t_One = vec4(1.0);
 `;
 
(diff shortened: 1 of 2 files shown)


src/BanjoKazooie/render.ts
```diff
@@ -262,21 +270,26 @@ void main() {
 
 #ifdef USE_TEXTURE
     t_Tex0 = Texture2D_N64(PP_SAMPLER_2D(u_Texture[0]), v_TexCoord.xy);
-    t_Tex1 = Texture2D_N64(PP_SAMPLER_2D(u_Texture[1]), v_TexCoord.zw);
+    #ifdef TWO_CYCLE
+        t_Tex1 = Texture2D_N64(PP_SAMPLER_2D(u_Texture[1]), v_TexCoord.zw);
+    #else
+        t_Tex1 = t_Tex0
```
PFedak, line 276: missing semicolon

src/BanjoKazooie/render.ts
```diff
@@ -62,6 +62,7 @@ varying vec4 v_Color;
 varying vec4 v_TexCoord;
 
 const vec4 t_Zero = vec4(0.0);
+const vec4 t_Half = vec4(0.5);
 const vec4 t_One = vec4(1.0);
```
magcius, line 66: This is pre-existing, so I don't expect a fix here, but just so people are aware: The 't_` was meant to mean "temporary", as in a local variable or a function parameter.

I chose this convention because GLSL has poor scoping rules, and it's too easy to accidentally rely on globals (uniforms or varyings) in places where unintended. I'm happy to phase it out if people don't find it helpful, but it helps if people are following the conventions. I think I typically use `c_` for global constant data.

src/BanjoKazooie/render.ts
```diff
@@ -292,6 +305,116 @@ ${this.generateAlphaTest()}
 }
 `;
     }
+
+    // Some debug methods to find code that uses interesting properties of the color combiner
+    public static usesTexelsInSecondCycle(comb: CombineParams): boolean {
+        for(let param of [comb.a1.a, comb.a1.b, comb.a1.c, comb.a1.d, comb.c1.a, comb.c1.b, comb.c1.c, comb.c1.d]) {
+            if(param === 1 || param === 2)
```
magcius, line 312: We have an enum for this: `CCMUX.TEXEL0`

src/BanjoKazooie/render.ts
RolandMunsil: Nah that seems like a good convention, I was just following the pattern of the other two. Changing them to `c_` would make sense

PR #338, opened 2020-09-06
Fine control camera hotkey

OverEngineeredCode:
In most engines, the default behaviour for noclip cameras is to slow the camera to from the set speed when the control key is pressed. Most of the time, the value is reduced to 1/10th. This is usually used to finely control the camera into small areas. This kind of tripped me up at first when I came across the site - because I was used to this hotkey working. So, I've added it using the same logic as the shift hotkey.

I've made this hotkey takes priority over the shift key, but I can easily change it if you think it should be the other way around.

## Example in Gary's Mod

If I use ```cl_showpos 1``` to show the velocity of the camera, you can see that the default speed is 500 Hu/S:

Normal noclip speed

But if I press shift while noclipping, the speed triples

Sprint noclip speed

And when pressing control it is divided by ten

Control noclip speed

## Problem with default hotkeys

Because of interference by the browser, for this hotkey to work without closing the tab or saving a copy (for example with Ctrl-W and Ctrl-S), you have to start moving with a movement key and then press the control key. I'll see if I can find a workaround that doesn't disrupt the default browser hotkeys too much.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ If you would like contribute, there is a Getting Started guide in the [Official 
 	- WASD/Arrow Keys: Move camera
 	- IJKL/Drag Mouse: Pan/tilt camera
 	- Shift: Increase camera speed
-	- Control: Reduce camera speed
+	- Backslash: Reduce camera speed
 	- Q/Page Down/Ctrl+Space: Move camera down
 	- E/Page Up/Space: Move camera up
 	- Scroll Wheel: Change camera speed
--- a/README.md
+++ b/README.md
@@ -26,6 +26,7 @@ If you would like contribute, there is a Getting Started guide in the [Official 
 	- WASD/Arrow Keys: Move camera
 	- IJKL/Drag Mouse: Pan/tilt camera
 	- Shift: Increase camera speed
+	- Control: Reduce camera speed
 	- Q/Page Down/Ctrl+Space: Move camera down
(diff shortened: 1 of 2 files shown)


magcius:
Hi, and thank you for the contribution! Yeah, this would be a great addition, but as you've noted, it tends to be a bit of a hassle with the browser keybindings, hence why I've excluded it so far. If you find an approach that prevents that, I would love this addition!

OverEngineeredCode:
I'm trying to write some logic in the InputManager that prevents the browser hotkeys that interfere with the WSAD keys. I'll do some experiments and then push it so you can tell me what you think. Hopefully, there's a way I can do this without outright overriding the browser keys.

OverEngineeredCode:
After some thought, I've remapped the slowdown key to backslash. I chose backslash because it's close to the shift key and so is just as convenient as control.

I tried to find a workaround to override browser keybinds, but it turns out that you can't do that for security reasons. While it is true that this makes the change less faithful to what most engines do, I think it could still be a useful feature to quickly slow down your camera without affecting the base speed.

If you want me to revisit this at some point in the future when more than two browsers support the keyboard locking API to hook to the control key properly, I am perfectly happy to do so.

magcius:
Backslash is an awkward key -- I'd prefer to just use the mouse wheel -- but I don't see any major harm from this, so merging.

PR #341, opened 2020-09-11
Moved game data sceneGroup files to src/Content

EntranceJew:
I tried to follow some basic principles when refactoring, roughly in order of
simplest to most complicated situations:
* Games that were self-contained were moved to Content/ without any changes
  if their `id` was identical
* Games that were self-contained that did not match their `id` were renamed
  and then moved to Content/
    * There is a small chance this will have broken pathName constants if
      they were identical
* Games that simply had a sceneGroup definition file in a folder among other
  sceneGroup files were moved with into a folder with no other code files
* Games that were based off already-completed games like mm3d or donkey kongs
  using metroid code were moved to the Content/ folder on their own
* Games that had stuff based off them in the same folder were moved to Content/
  with all the corresponding code, leaving the games based off them relative,
  from their appropriate place in Content/
* Games that existed in folders named unrelated to themselves had the content
  of that folder name left at the root of src/ while all the game scene files
  were relocated to Content/
* GTA/GTA3 is too complicated so I didn't touch it
* Jasper said not to touch the SuperMarioGalaxy folder for now, so I didn't


magcius:
Ultimately, this is a large change without many large benefits, and I think it makes several things worse. The renames from e.g. "snap" to "PokemonSnap" was very much intentional, and meant for readability. To be clear; the new naming convention I've been using in most projects is PascalCaseFilenames.ts, with the Scenes files being named Scenes_BlahBlahBlah.ts.

I like the idea of the Content/ folder, though things can be messy. If we're going to do this, I would like to move everything into Content/ but keep the current folder structure the same, except maybe move *some* things into Common/, like gx/ moving to Common/GX/. So something like the following folder structure for src/:

* assembly/
* assets/
* Common/
* Content/
* gfx/
* tools/
* vendor/

Some of those should probably be renamed, but we can do that later.

Ultimately, I'm going to reject this in its current form. If you have any trouble navigating the codebase, please let me know.

PR #343, opened 2020-09-19
Add experimental support for Beetle Adventure Racing

RolandMunsil:
Feeling like taking a break from this for a bit, so I figured I'd commit what I have so far as experimental support. Wasn't sure how strict your coding standards are so I may have left this code messier than you'd like, but obviously I'm happy to change whatever's needed to bring it up to snuff 🙂

Also I don't know if it matters, but the main reasoning behind this being experimental is:
 * Fair bit of alpha blending/depth comparison issues
 * Animated models not implemented
 * Billboarded objects don't rotate towards camera


src/BeetleAdventureRacing/Scenes.ts
```diff
@@ -0,0 +1,295 @@
...
+        // Render
+        if(this.uvenRenderer !== null)
+            this.uvenRenderer.prepareToRender(device, renderInstManager, viewerInput);
+
+        // Final setup
+        this.renderHelper.renderInstManager.popTemplateRenderInst();       
+        this.renderHelper.prepareToRender(device, hostAccessPass);
+    }
+
+    //TODO-ASK: how does this work? what is a pass? what is the host access pass? what is the return value?
```
magcius, line 146: You mostly don't need to care about these questions. GfxHostAccessPass is how you upload data to the GPU (the name is likely confusing, the name and partial API design weretaken from this old proposal https://github.com/gpuweb/gpuweb/issues/45 ).

Passes themselves are how you draw things -- you draw onto a render target, and "resolve" that render target to a texture which is ultimately displayed. You don't have to care too deeply about the resolve; it's just how antialiasing works. If you want, think of it as rendering a giant version of the picture, and then the "resolve" resizes it with antialiasing turned on.

The return value here is a bit weird, as you return a render pass to be "resolved" to the screen. I never liked this, so a few months ago I added support so you can now do:

```ts
const passRenderer = this.renderTarget.createRenderPass(device, viewerInput.viewport, this.renderPassDescriptor, viewerInput.onscreenTexture);
executeOnPass(renderInstManager, device, passRenderer, 0);
device.submitPass(passRenderer);

return null;
```

... to be a bit more explicit. You're creating a new render pass, rendering a bunch of stuff, and at the end of everything, going to resolve all of that information to the "onscreen texture" (that's what will be displayed by the browser). And then the "return null;" just means "I'm not returning a pass for you to resolve". Basically, if you return a render pass, the viewer code that calls `render` [just does that for you automatically](https://github.com/magcius/noclip.website/blob/98ac2bd331ba3ecff25983a335986c51d6d93718/src/viewer.ts#L162-L174).

Sorry if this made things more confusing. The code you have here is fine. If you have any more questions, feel free to ask.

src/BeetleAdventureRacing/Scenes.ts
```diff
@@ -0,0 +1,295 @@
...
+    constructor(public sceneIndex: number | null, public name: string, public uvtrIndex: number | null = null) {
+        if (this.sceneIndex !== null) {
+            this.id = "sc" + this.sceneIndex.toString();
+        } else {
+            this.id = "tr" + this.uvtrIndex;
+        }       
+    }
+
+    public async createScene(device: GfxDevice, context: SceneContext): Promise<SceneGfx> {
+        // TODO maybe this and RendererStore should be global variables instead of passing them around everywhere?
```
magcius, line 184: Please don't make them globals, as that might break if the user loads multiple scenes in succession. The `dataShare` is the mechanism to effectively create globals. If you don't like passing 20 things around everywhere, I sometimes create a `class BeetleAdventureGlobals {}` object to hang things off of, and pass *that* around everywhere.

src/BeetleAdventureRacing/MaterialRenderer.ts
```diff
@@ -0,0 +1,414 @@
...
+
+        if(this.DEBUG_shouldSkip()) {
+            this.isTextured = false;
+            return;
+        }
+
+        // Rendering config
+        let stateFlags;
+        let otherModeLRenderMode: number;
+        ({ stateFlags, otherModeLRenderMode } = this.translateRenderOptions());
```
magcius, line 59: You can just do `let { stateFlags, otherModeLRenderMode } = this.translateRenderOptions();`

magcius (approved):
Sorry for the late review on this.

My general strategy for code maintenance is to let game authors define their own coding style. Any changes to shared code, I will scrutinize much more closely. And things like leaks that can affect site stability, I'm a bit more cautious of those. But otherwise, go hog wild, it's your code, I'm not gonna tell you what to do :) I left a few comments, but none of them are requirements for merge, just tiny cleanups here and there that you might enjoy.

To actually merge this, I will need a filesystem.bin. You can just DM that to me on Discord and I'll put it up. But otherwise, I think it's good to go.

src/BeetleAdventureRacing/Scenes.ts
RolandMunsil: No, this is helping! Admittedly I still have a lot of questions heh but it does make more sense now 🙂

PR #345, opened 2020-09-27
viewer: Increase default FoV from 45 to 60 degrees

VelocityRa:
Low FoV values such as 45 are common in console games where the TV
screen is usually quite far, but not so in Desktop/mobile scenarioes
where noclip.website is most commonly used for.

Values of 80-90 are much more common in these cases, I picked 60 since
it's around the middle (also it divided evenly).

---
The exact value is of course open to bikeshedding (should probably be even more?), but I do believe it's necessary to increase it.


magcius:
So this is potentially a giant change, but let's try it out. I'm feeling bold this morning. It might eventually make sense for scenes to define their own default FoV settings (maybe as part of default savestates?)

PR #362, opened 2020-10-22
Fix incorrect alpha parameter in F3DEX_Program

RolandMunsil:
First alpha value was set to `combAlpha` for all alpha parameters, but for the multiply parameter it's supposed to be LOD_FRACTION (See [here](https://github.com/gonetz/GLideN64/blob/master/src/Combiner.cpp#L63)). Noticed because it was causing some textures in BAR to have the incorrect transparency.

### DIFF
--- a/src/BanjoKazooie/render.ts
+++ b/src/BanjoKazooie/render.ts
@@ -219,8 +219,9 @@ void main() {
             'v_Color.a', 'u_EnvColor.a', '1.0', '0.0'
         ];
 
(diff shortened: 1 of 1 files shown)


magcius:
I care somewhat about LOD_FRACTION. It matters for certain games with custom mipmaps, e.g. that's how Super Mario 64 does the fade from Peach to Bowser in that hallway, and there's a similar effect done for the rainbow floor in Paper Mario 64, but that codebase doesn't quite use the RDP shader yet, and when I looked into it, calculating LOD_FRACTION accurately seemed tough.

PR #371, opened 2020-11-05
Default actor animations for Metroid Prime 1 and 2

jackoalan:
* Loaders for animation-related resources
* Meta animation and transition node implementations
* CPU skinning for MP1
* Loads existing envelope matrix sets for MP2
* "Show All Actors" and "Suit Model" render hacks exposed to user

### DIFF
--- a/src/gx/gx_displaylist.ts
+++ b/src/gx/gx_displaylist.ts
@@ -470,7 +470,8 @@ function translateVatLayout(vatFormat: GX_VtxAttrFmt[], vcd: GX_VtxDesc[]): VatL
             srcVertexSize += getAttributeByteSizeRaw(vtxAttrib, vtxAttrFmt);
         else if (vtxAttrDesc.type === GX.AttrType.INDEX8)
             srcVertexSize += 1 * srcIndexComponentCount;
-        else if (vtxAttrDesc.type === GX.AttrType.INDEX16)
+        else if (vtxAttrDesc.type === GX.AttrType.INDEX16 ||
+                 vtxAttrDesc.type === GX.AttrType.UNRESOLVED_INDEX16)
             srcVertexSize += 2 * srcIndexComponentCount;
     }
 
@@ -530,7 +531,11 @@ export function compileLoadedVertexLayout(vat: GX_VtxAttrFmt[][], vcd: GX_VtxDes
         let fieldFormat: GfxFormat;
         let fieldCompOffset: number = 0;
 
-        if (isVtxAttribTexMtxIdx(vtxAttrib)) {
+        if (vtxAttrDesc.type === GX.AttrType.UNRESOLVED_INDEX16) {
+            const attrInput = getAttrInputForAttr(vtxAttrib);
+            input = allocateVertexInput(attrInput, GfxFormat.U16_R);
+            fieldFormat = input.format;
+        } else if (isVtxAttribTexMtxIdx(vtxAttrib)) {
             // Allocate the base if it doesn't already exist.
             const attrInput = (vtxAttrib < GX.Attr.TEX4MTXIDX) ? VertexAttributeInput.TEX0123MTXIDX : VertexAttributeInput.TEX4567MTXIDX;
             input = allocateVertexInput(attrInput, inputFormat);
@@ -641,6 +646,12 @@ function generateRunVertices(loadedVertexLayout: LoadedVertexLayout, vatLayout: 
             const littleEndian = (getSystemEndianness() === Endianness.LITTLE_ENDIAN);
             const dstOffs = `dstVertexDataOffs + ${offs}`;
             return `dstVertexDataView.setFloat32(${dstOffs}, ${value}, ${littleEndian})`;
+        }
+
+        function compileWriteOneComponentU16(offs: number, value: string): string {
+            const littleEndian = (getSystemEndianness() === Endianness.LITTLE_ENDIAN);
+            const dstOffs = `dstVertexDataOffs + ${offs}`;
+            return `dstVertexDataView.setUint16(${dstOffs}, ${value}, ${littleEndian})`;
         }
 
         function compileWriteOneComponentU8Norm(offs: number, value: string): string {
@@ -810,6 +821,12 @@ function generateRunVertices(loadedVertexLayout: LoadedVertexLayout, vatLayout: 
             return compileAttribIndex(compileVtxArrayViewName(vtxAttrib), `dlView.getUint8(drawCallIdx)`, 1);
         case GX.AttrType.INDEX16:
             return compileAttribIndex(compileVtxArrayViewName(vtxAttrib), `dlView.getUint16(drawCallIdx)`, 2);
+        case GX.AttrType.UNRESOLVED_INDEX16:
+            return `
+    // UNRESOLVED ${getAttrName(vtxAttrib)}
+    ${compileWriteOneComponentU16(dstBaseOffs, `dlView.getUint16(drawCallIdx)`)};
+    drawCallIdx += 2;
+`;
         default:
             throw "whoops";
         }
--- a/src/metroid_prime/ancs.ts
+++ b/src/metroid_prime/ancs.ts
@@ -1,23 +1,28 @@
-
 // Implements Retro's ANCS format as seen in Metroid Prime 1.
 
-import { assert } from "../util";
+import { assert } from '../util';
 
-import { ResourceSystem } from "./resource";
-import { CMDL } from "./cmdl"
-import { InputStream } from "./stream";
+import { ResourceSystem } from './resource';
+import { CMDL } from './cmdl';
+import { InputStream } from './stream';
 
-// minimal implementation of ANCS
+import { AdditiveAnimation, Animation, AnimationSet, CreateMetaAnim, CreateMetaTrans, HalfTransition, Transition, TransitionDatabase } from './animation/meta_nodes';
+import { CINF } from './cinf';
+
+// minimal implementation of character set entry
 export interface MetroidCharacter {
     charID: number;
     name: string;
     model: CMDL | null;
+    skel: CINF | null;
     skinID: string;
     skelID: string;
+    animNames: string[];
 }
 
 export interface ANCS {
-    characters: MetroidCharacter[];
+    characterSet: MetroidCharacter[];
+    animationSet: AnimationSet;
 }
 
 export function parse(stream: InputStream, resourceSystem: ResourceSystem, assetID: string): ANCS {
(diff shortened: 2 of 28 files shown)


src/gl-matrix-ext.ts
```diff
@@ -0,0 +1,66 @@
...
+    /**
+     * Scales the mat3 by the dimensions in the given vec3
+     *
+     * @param {mat3} out the receiving matrix
+     * @param {ReadonlyMat3} a the matrix to rotate
+     * @param {ReadonlyVec3} v the vec2 to scale the matrix by
+     * @returns {mat3} out
+     **/
+    static scale3(out: mat3, a: ReadonlyMat3, v: ReadonlyVec3): mat3 {
+        var x = v[0],
```
magcius, line 17: I just prefer not to use mat3 in general; it doesn't save us much over mat4.

src/gx/gx_displaylist.ts
```diff
@@ -464,9 +464,12 @@ function translateVatLayout(vatFormat: GX_VtxAttrFmt[], vcd: GX_VtxDesc[]): VatL
 
         const srcIndexComponentCount = getIndexNumComponents(vtxAttrib, vtxAttrFmt);
 
-        // MTXIDX entries can only be DIRECT if they exist.
+        // MTXIDX entries can only be DIRECT or GENERATED_MTXIDX if they exist.
```
magcius, line 467: This is the big point of contention. I'll have to think about how to better do this better.

I honestly probably wouldn't be opposed to building GPU-side fine-skinning tech. I know beholdnec has wanted this for SFA as well, and I think it's not an unreasonable request. We can look into building this together.

Sorry it's taken me so long to come to this conclusion and ask you to redo this twice. That is entirely my fault.

src/metroid_prime/animation/base_reader.ts
```diff
@@ -0,0 +1,71 @@
...
+    abstract GetPerSegmentData(indices: number[], time?: CharAnimTime): PerSegmentData[];
+
+    abstract SetPhase(phase: number): void;
+
+    abstract Clone(): IAnimReader;
+
+    /**
+     * Periodically called to cull out completed transition nodes
+     */
+    Simplified(): IAnimReader | null {
```
magcius, line 68: Please add visibility accessors -- `public` in this case.

src/metroid_prime/animation/char_anim_time.ts
```diff
@@ -0,0 +1,223 @@
...
+
+    EqualsZero(): boolean {
+        switch (this.state) {
+            case AnimTimeState.ZeroIncreasing:
+            case AnimTimeState.ZeroSteady:
+            case AnimTimeState.ZeroDecreasing:
+                return true;
+        }
+
+        return this.time == 0.0;
```
magcius, line 35: try to use triple-equality in all cases

src/metroid_prime/animation/data_source.ts
```diff
@@ -0,0 +1,291 @@
...
+import { nArray } from "../../util";
+
+class KeyStorage {
+    scaleKeys: ReadonlyVec3[];
+    rotationKeys: ReadonlyQuat[];
+    translationKeys: ReadonlyVec3[];
+
+    constructor(stream: InputStream, private frameCount: number, mp2: boolean) {
+        if (mp2)
+            this.scaleKeys = nArray(stream.readUint32(), () => stream.readVec3(vec3.create()));
```
magcius, line 17: This will leave scaleKeys uninitialized if mp2 is not passed in. In that case, I'd decorate scaleKeys as `private scaleKeys: ReadonlyVec3[] | null = null;`

src/metroid_prime/animation/data_source.ts
```diff
@@ -0,0 +1,291 @@
...
+        this.rotationKeys = nArray(stream.readUint32(), () => stream.readQuat(quat.create()));
+        this.translationKeys = nArray(stream.readUint32(), () => stream.readVec3(vec3.create()));
+    }
+
+    GetScale(frameIdx: number, scaleIdx: number): ReadonlyVec3 {
+        return this.scaleKeys[this.frameCount * scaleIdx + Math.min(frameIdx, this.frameCount - 1)];
```
magcius, line 23: ... and use `const scaleKeys = this.scaleKeys!` -- calling GetScale would be an object with null scaleKeys is an error, but that's fine, it is already anyway :)

src/metroid_prime/mrea.ts
```diff
@@ -561,7 +578,7 @@ function parseWorldModels_MP3(stream: InputStream, worldModelCount: number, wobj
     return worldModels;
 }
 
-function parseSurfaces(stream: InputStream, surfaceCount: number, sectionIndex: number, posSectionOffs: number, nrmSectionOffs: number, clrSectionOffs: number, uvfSectionOffs: number, uvsSectionOffs: number | null, sectionOffsTable: number[], worldModelIndex: number, materialSet: MaterialSet, isEchoes: boolean): [Surface[], number] {
+function parseSurfaces(stream: InputStream, surfaceCount: number, sectionIndex: number, posSectionOffs: number, nrmSectionOffs: number, clrSectionOffs: number, uvfSectionOffs: number, uvsSectionOffs: number | null, sectionOffsTable: number[], worldModelIndex: number, materialSet: MaterialSet, isEchoes: boolean, cskr?: CSKR): [Surface[], number] {
```
magcius, line 581: Sorry for the mess that is `parseSurfaces`. It could probably do with a bigger refactor.

src/metroid_prime/mrea.ts
```diff
@@ -598,8 +615,10 @@ function parseSurfaces(stream: InputStream, surfaceCount: number, sectionIndex:
         const normalY = stream.readFloat32();
         const normalZ = stream.readFloat32();
 
+        let sourceEnvelopeIdx: number | undefined = undefined;
```
magcius, line 618: I try to use null instead of undefined where possible now.

src/metroid_prime/mrea.ts
```diff
@@ -656,7 +720,158 @@ function parseSurfaces(stream: InputStream, surfaceCount: number, sectionIndex:
         vtxArrays[GX.Attr.TEX6] = { buffer: stream.getBuffer(), offs: uvfSectionOffs, stride: getAttributeByteSize(fmtVat, GX.Attr.TEX6) };
         vtxArrays[GX.Attr.TEX7] = { buffer: stream.getBuffer(), offs: uvfSectionOffs, stride: getAttributeByteSize(fmtVat, GX.Attr.TEX7) };
 
-        const loadedVertexData = vtxLoader.runVertices(vtxArrays, dlData);
+        let loadedVertexData = vtxLoader.runVertices(vtxArrays, dlData);
+
+        if (generateMtxIdxs) {
+            const littleEndian = getSystemEndianness() === Endianness.LITTLE_ENDIAN;
```
magcius, line 726: lol, noclip will break incredibly badly on big-endian devices, I'm quite sure.

src/metroid_prime/render.ts
```diff
@@ -193,11 +261,13 @@ class MaterialGroupInstance {
...
 
-        const layer = this.material.isTransparent ? GfxRendererLayer.TRANSLUCENT : GfxRendererLayer.OPAQUE;
+        const layer = this.material.isDepthSorted ? GfxRendererLayer.TRANSLUCENT : GfxRendererLayer.OPAQUE;
         renderInst.sortKey = makeSortKey(layer, this.materialHelper.programKey);
+        if (this.material.isDepthSorted)
+            renderInst.sortKey = setSortKeyBias(renderInst.sortKey, this.material.sortBias);
     }
 
-    public prepareToRender(renderInstManager: GfxRenderInstManager, viewerInput: Viewer.ViewerRenderInput, modelMatrix: mat4, isSkybox: boolean, actorLights: ActorLights | null, worldAmbientColor: Color): void {
+    public prepareToRender(renderInstManager: GfxRenderInstManager, viewerInput: Viewer.ViewerRenderInput, modelMatrix: mat4, isSkybox: boolean, actorLights: ActorLights | null, worldAmbientColor: Color, envelopeModelMatrices?: (mat4|null)[]): void {
```
magcius, line 270: This can be `envelopeModelMatrices: (mat4 | null)[] = []`, right?

src/metroid_prime/scenes.ts
```diff
@@ -137,21 +160,32 @@ export class RetroSceneRenderer implements Viewer.SceneGfx {
...
 
     public async createScene(device: GfxDevice, context: SceneContext): Promise<Viewer.SceneGfx> {
         const dataFetcher = context.dataFetcher;
         const levelPak = PAK.parse(await dataFetcher.fetchData(`metroid_prime/${this.filename}`), this.gameCompressionMethod);
-        const resourceSystem = new ResourceSystem([levelPak]);
+        const resourceSystem = new ResourceSystem(this.game, [levelPak]);
+
+        // Preload all ANCS resources to associate models with skins
+        for (const entry of levelPak.resourceTable.values()) {
+            if (entry.fourCC === 'ANCS') {
```
magcius, line 185: Hmm, won't this be expensive to load everything, even for rooms that don't use certain actors? Can we not do this more lazily?

src/metroid_prime/script.ts
```diff
@@ -113,22 +117,37 @@ export class Entity {
     public readProperty_MP2(stream: InputStream, resourceSystem: ResourceSystem, propertyID: number) {
     }
 
-    public getRenderModel(): CMDL | null {
+    public getRenderModel(resourceSystem: ResourceSystem): [CMDL | null, AnimationData?] {
```
magcius, line 120: I might prefer

```ts
interface RenderModel {
    cmdl: CMDL;
    animationData : AnimationData | null;
}
```

then you can do `return { cmdl, animationData };`, which I find a bit cleaner than array-based returns.

src/metroid_prime/stream.ts
```diff
@@ -46,4 +47,19 @@ export class InputStream {
...
+        v[1] = this.readFloat32();
+        v[2] = this.readFloat32();
+        return v;
+    }
+
+    public readQuat(q: quat): quat {
+        q[1] = this.readFloat32();
+        q[2] = this.readFloat32();
+        q[3] = this.readFloat32();
+        q[0] = this.readFloat32();
```
magcius, line 62: this sure is a packing choice

magcius:
Incredibly sorry about the delay in reviewing this.

General note on reviews: contributors are allowed to define their own styles and conventions for their game subtrees. I will merge this PR even if you take none of my suggestions. I'm just giving examples of how I've solved problems in the past, and some of the conventions of the noclip codebase in general.

The bigger exception to this is shared tools: the GX display list interpreter is important to me that it remains flexible and performance, and I don't want to weigh it down with too many options. At the same time, I understand that it imposes a structure that not all GameCube games follow, and I want to make sure it remains helpful. We've chatted a bit about possible architectures for how do to this, and I'm confident we can find a solution both of us are happier with. I think we have similar goals: try not to go too crazy on data processing, cut down on draw calls, and make sure that we aren't too invasive to the GX helpers.

src/metroid_prime/scenes.ts
jackoalan: This is a weird one to solve. Some areas may present a skinned CMDL as a static one where a later area presents it skinned. In this situation, I'd need to reload an already-loaded model.

src/gx/gx_displaylist.ts
jackoalan: No matter if the position indices are extracted in-band (W coordinate to be relocated by the game loader) or out-of-band (as separate `LoadedVertexData` variable), this will require similarly invasive changes to the display list loader.

I would be willing to proceed with either approach. I personally favor the in-band relocation approach because the vertex buffer W coordinates will need to be patched either way.

src/metroid_prime/ancs.ts
```diff
@@ -34,37 +47,41 @@ export function parse(stream: InputStream, resourceSystem: ResourceSystem, asset
...
         for (let nameIdx = 0; nameIdx < numAnimNames; nameIdx++) {
             const animID = stream.readUint32();
             if (charVersion < 10) {
                 const unk = stream.readString();
             }
-            const animName = stream.readString();
+            animNames[nameIdx] = stream.readString();
         }
 
+        const char: MetroidCharacter = {charID, name, model, skel: skel ? skel : undefined, skinID, skelID, animNames};
```
jackoalan, line 65: Use null as-is rather than converting to undefined

PR #383, opened 2020-12-31
Improved MKDD. Objects, BCA, rotation fix and more

Wexos:
First time writing in TypeScript... I have changed some stuff for MKDD.
* Added support for BCA animations (probably bad implementation).
* Implemented BOL rotation correctly (very wrong before, even though it worked most of the times). I also added reading for object settings.
* Changed how files are loaded from the RARC course. It uses a more similar way of reading data which the original game does. It also reads all supported animations of both course BMD and skybox BMD.
* Changed how objects are loaded. They now must specify both the BMD name and what animations to load.
* Added a lot more objects. The object loading is simplified, only the BMD and in some cases, animation files, are loaded. Nothing else has been done (no moving etc). The file names are hardcoded in a giant switch statement. I have also added objects which doesn't use any models to the switch, but they currently do nothing. Also, objects which are stored in MRAM.arc are currently not loaded. Don't really like the giant switch, but until better object replication is done, this is all that's needed I guess.
* Added two extra courses, Award.arc and Luigi2.arc.

There is one problem, which also was a problem before. I have not found a way to bind multiple BTK animations to the same BMD. The course BMD can be animated by three different BTK files. As far as I understand, the current implementation does not support this. This means currently, only the last loaded course BTK file be animated.

### DIFF
--- a/src/Common/JSYSTEM/J3D/J3DGraphAnimator.ts
+++ b/src/Common/JSYSTEM/J3D/J3DGraphAnimator.ts
@@ -1,7 +1,7 @@
 
 // Animation support.
 
-import { AnimationTrack, AnimationKeyframe, LoopMode, VAF1, TRK1, TRK1AnimationEntry, calcTexMtx_Maya, calcTexMtx_Basic, TTK1, TTK1AnimationEntry, TPT1AnimationEntry, TPT1, Joint, JointTransformInfo, ANK1, ANK1JointAnimationEntry, J3DLoadFlags } from './J3DLoader';
+import { AnimationTrack, AnimationKeyframe, LoopMode, VAF1, TRK1, TRK1AnimationEntry, calcTexMtx_Maya, calcTexMtx_Basic, TTK1, TTK1AnimationEntry, TPT1AnimationEntry, TPT1, Joint, JointTransformInfo, ANK1, ANK1JointAnimationEntry, J3DLoadFlags, ANF1JointAnimationEntry } from './J3DLoader';
 import { assertExists } from '../../../util';
 import { Color } from '../../../Color';
 import { J3DModelInstance, JointMatrixCalcNoAnm, MaterialInstance, J3DModelData, ShapeInstanceState } from './J3DGraphBase';
@@ -339,6 +339,9 @@ export function calcJointMatrixMayaSSC(dst: mat4, parentScale: ReadonlyVec3): vo
 }
 
 export function calcJointMatrixFromTransform(dst: mat4, transform: JointTransformInfo, loadFlags: J3DLoadFlags, jnt1: Joint, shapeInstanceState: ShapeInstanceState): void {
+    
+    
+    
     mat4.fromQuat(dst, transform.rotation);
     setMatrixTranslation(dst, transform.translation);
     mat4.scale(dst, dst, transform.scale);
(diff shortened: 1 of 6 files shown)


src/Common/JSYSTEM/J3D/J3DGraphAnimator.ts
```diff
@@ -412,3 +415,37 @@ export function entryJointAnimator(modelInstance: J3DModelInstance, ank1: ANK1,
 export function removeJointAnimator(modelInstance: J3DModelInstance, ank1: ANK1): void {
     modelInstance.jointMatrixCalc = new JointMatrixCalcNoAnm();
 }
+
+const anf1ScratchQuat = quat.create();
```
magcius, line 419: You should be able to reuse the scratchQuat from above.

src/Common/JSYSTEM/J3D/J3DGraphAnimator.ts
```diff
@@ -412,3 +415,37 @@ export function entryJointAnimator(modelInstance: J3DModelInstance, ank1: ANK1,
...
+export function calcANF1JointAnimationTransform(dst: JointTransformInfo, entry: ANF1JointAnimationEntry, animFrame: number, animFrame1: number): void {
+
+    const sample = (frames: number[]) => {
```
magcius, line 422: Can you move this out of this function? Closures (nested functions) cause a heap object to be created on every function call.

src/Common/JSYSTEM/J3D/J3DGraphAnimator.ts
```diff
@@ -339,6 +339,9 @@ export function calcJointMatrixMayaSSC(dst: mat4, parentScale: ReadonlyVec3): vo
 }
 
 export function calcJointMatrixFromTransform(dst: mat4, transform: JointTransformInfo, loadFlags: J3DLoadFlags, jnt1: Joint, shapeInstanceState: ShapeInstanceState): void {
+
```
magcius, line 342: Remove the added whitespace please.

src/Common/JSYSTEM/J3D/J3DGraphSimple.ts
```diff
@@ -51,8 +51,35 @@ class JointMatrixCalcANK1 {
...
+        }
+
+        const loadFlags = modelData.bmd.inf1.loadFlags;
+        calcJointMatrixFromTransform(dst, transform, loadFlags, jnt1, shapeInstanceState);
+
+        vec3.copy(shapeInstanceState.parentScale, transform.scale);
+    }
+}
+
+const anf1ScratchTransform = new JointTransformInfo();
```
magcius, line 67: Should be possible to reuse scratchTransform for both.

src/j3d/mkdd_scenes.ts
```diff
@@ -131,18 +132,51 @@ function parseBOL(buffer: ArrayBufferSlice): BOL {
...
         const routeId = view.getInt16(objectTableIdx + 0x26);
+        const settings = [];
 
+        for (let i = 0; i < 8; i++) {
+            settings.push(view.getUint16(objectTableIdx + 0x30 + i * 2));
+        }
+
+        // Create rotation
+        const forward: vec3 = [forwardX, forwardY, forwardZ];
+        const up: vec3 = [upX, upY, upZ];
```
magcius, line 151: Please use vec3.fromValues() instead of creating arrays with [].

magcius (commented):
> I have not found a way to bind multiple BTK animations to the same BMD. The course BMD can be animated by three different BTK files. As far as I understand, the current implementation does not support this.

You can hand-call bindTTK1MaterialInstance / bindTTK1 yourself. The texMtxCalc callback is on the material instance; the bindTTK1 helper that's part of J3DModelInstanceSimple won't handle complex cases like this, you have to write a bit of code yourself to do it.

Wexos:
All issues should now be fixed.

PR #384, opened 2021-01-12
Ota king impl

cbondurant:
This implements more features of OtaKing, such as adding his arms, and having him face towards the camera as he does in the game. Im not done working on this character, as I intend to add in his firing of coconuts and fireballs at some point. But this is more than enough to go ahead and push I think.

### DIFF
--- a/src/SuperMarioGalaxy/Actors/Enemy.ts
+++ b/src/SuperMarioGalaxy/Actors/Enemy.ts
@@ -5082,9 +5082,9 @@ class CocoNutBall extends LiveActor<CocoNutBallNrv> {
         this.initModelManagerWithAnm(sceneObjHolder, "CocoNut");
         connectToSceneNoSilhouettedMapObjStrongLight(sceneObjHolder, this);
         initLightCtrl(sceneObjHolder, this);
-        this.initHitSensor()
-        addHitSensor(sceneObjHolder, this, "body", HitSensorType.CocoNut, 8,40,[0,0,0]);
-        addHitSensor(sceneObjHolder, this, "bind", HitSensorType.CocoNut, 8,40,[0,0,0]);
+        this.initHitSensor();
+        addHitSensor(sceneObjHolder, this, "body", HitSensorType.CocoNut, 8, 40, [0,0,0]);
+        addHitSensor(sceneObjHolder, this, "bind", HitSensorType.CocoNut, 8, 40, [0,0,0]);
         this.initBinder(60, 0, 0);
         this.initEffectKeeper(sceneObjHolder, "CocoNut");
         // initStarPointerTarget
@@ -5121,7 +5121,6 @@ export class OtaKingLongFoot extends PartsModel<OtaKingLongFootNrv>{
         showModel(this);
         startBck(this,"Wait");
         setBckFrameAtRandom(this);
-        //setBckFrameAndStop(this, 0)
     }
 
     protected control(sceneObjHolder: SceneObjHolder, viewerInput: Viewer.ViewerRenderInput): void {
@@ -5165,7 +5164,6 @@ export class OtaKing extends LiveActor<OtaKingNrv> {
 
             this.dirToPlayer(sceneObjHolder);
             this.tryThrowCoconutOrFireBallIfWait(0x4b);
-
         }
     }
 
@@ -5202,24 +5200,24 @@ export class OtaKing extends LiveActor<OtaKingNrv> {
     private tryThrowCoconutOrFireBallIfWait(step: number): boolean {
         if (isGreaterStep(this, step) || !this.isValidThrowCoconut()){
             if (isGreaterStep(this, step << 1) || !this.isValidThrowFireball()){
-                return false
+                return false;
             }else{
                 this.setNerve(OtaKingNrv.ThrowFireBall);
-                return false
+                return false;
             }
         }else{
             this.setNerve(OtaKingNrv.ThrowCocoNut);
-            return true
+            return true;
         }
     }
 
     private isValidThrowFireball(): boolean {
         //throw new Error('Method not implemented.');
-        return false
+        return false;
     }
 
     private isValidThrowCoconut(): boolean {
-        return false
+        return false;
     }
 
     private startBckWaitIfNotPlaying(){
@@ -5232,21 +5230,21 @@ export class OtaKing extends LiveActor<OtaKingNrv> {
     public initLongFoot(sceneObjHolder:SceneObjHolder, infoIter: JMapInfoIter){
         this.longFeet = new Array<OtaKingLongFoot>();
         if (!this.isLv2){
-            let stepOffset = [3,8,0xE,0x12];
+            let stepOffset = [3, 8, 0xE, 0x12];
             for (let i = 0; i < 4; i++){
                 let longFoot = new OtaKingLongFoot(sceneObjHolder, infoIter, this, stepOffset[i]);
                 this.longFeet[i] = longFoot;
                 vec3.set(longFoot.scale, 1.0, 0.9, 0.8);
             }
 
-            vec3.add(this.longFeet[0].translation, vec3.set(scratchVec3a, 735.0, 80.0, -55.0), this.translation)
-            vec3.set(this.longFeet[0].rotation, -9*MathConstants.DEG_TO_RAD, 266*MathConstants.DEG_TO_RAD, 0.0)
-            vec3.add(this.longFeet[1].translation, vec3.set(scratchVec3a, -959.0, 130.0, 0.0), this.translation)
-            vec3.set(this.longFeet[1].rotation, 0.0,107*MathConstants.DEG_TO_RAD,14.0*MathConstants.DEG_TO_RAD)
-            vec3.add(this.longFeet[2].translation, vec3.set(scratchVec3a,  0.0, 43.0, 884.0), this.translation)
-            vec3.set(this.longFeet[2].rotation, -8*MathConstants.DEG_TO_RAD,159*MathConstants.DEG_TO_RAD,0.0)
-            vec3.add(this.longFeet[3].translation, vec3.set(scratchVec3a, 55.0, 55.0, -896.0), this.translation)
-            vec3.set(this.longFeet[3].rotation, -9*MathConstants.DEG_TO_RAD,-14*MathConstants.DEG_TO_RAD,0.0)
+            vec3.add(this.longFeet[0].translation, vec3.set(scratchVec3a, 735.0, 80.0, -55.0), this.translation);
+            vec3.set(this.longFeet[0].rotation, -9*MathConstants.DEG_TO_RAD, 266*MathConstants.DEG_TO_RAD, 0.0);
+            vec3.add(this.longFeet[1].translation, vec3.set(scratchVec3a, -959.0, 130.0, 0.0), this.translation);
+            vec3.set(this.longFeet[1].rotation, 0.0,107*MathConstants.DEG_TO_RAD,14.0*MathConstants.DEG_TO_RAD);
+            vec3.add(this.longFeet[2].translation, vec3.set(scratchVec3a,  0.0, 43.0, 884.0), this.translation);
+            vec3.set(this.longFeet[2].rotation, -8*MathConstants.DEG_TO_RAD,159*MathConstants.DEG_TO_RAD,0.0);
+            vec3.add(this.longFeet[3].translation, vec3.set(scratchVec3a, 55.0, 55.0, -896.0), this.translation);
+            vec3.set(this.longFeet[3].rotation, -9*MathConstants.DEG_TO_RAD,-14*MathConstants.DEG_TO_RAD,0.0);
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/SuperMarioGalaxy/Actors/Enemy.ts
```diff
@@ -5053,3 +5054,223 @@ export class Petari extends LiveActor<PetariNrv> {
         return true;
     }
 }
+
+enum CocoNutBallNrv { Throw };
```
magcius, line 5058: const enum

src/SuperMarioGalaxy/Actors/Enemy.ts
```diff
@@ -5053,3 +5054,223 @@ export class Petari extends LiveActor<PetariNrv> {
...
+        this.initHitSensor();
+        addHitSensor(sceneObjHolder, this, "body", HitSensorType.CocoNut, 8, 40, [0,0,0]);
+        addHitSensor(sceneObjHolder, this, "bind", HitSensorType.CocoNut, 8, 40, [0,0,0]);
+        this.initBinder(60, 0, 0);
+        this.initEffectKeeper(sceneObjHolder, "CocoNut");
+        // initStarPointerTarget
+        initShadowVolumeCylinder(sceneObjHolder, this, 60);
+        // invalidateClipping
+        this.initNerve(CocoNutBallNrv.Throw);
+        this.makeActorDead;
```
magcius, line 5076: You forgot to call the method

src/SuperMarioGalaxy/Actors/Enemy.ts
```diff
@@ -5053,3 +5054,223 @@ export class Petari extends LiveActor<PetariNrv> {
...
+        super.updateSpine(sceneObjHolder, currentNerve, deltaTimeFrames);
+        if (currentNerve === CocoNutBallNrv.Throw){
+            if (isFirstStep(this)){
+                startBck(this, "SpinX");
+                emitEffect(sceneObjHolder, this, "CocoNutLight");
+            }
+            this.processApproachToPlayer();
+        }
+    }
+    processApproachToPlayer() {
```
magcius, line 5089: private

Also won't this always throw?

src/SuperMarioGalaxy/Actors/Enemy.ts
```diff
@@ -5053,3 +5054,223 @@ export class Petari extends LiveActor<PetariNrv> {
...
+    }
+}
+
+enum OtaKingLongFootNrv { Wait }
+
+export class OtaKingLongFoot extends PartsModel<OtaKingLongFootNrv>{
+    private stepOffset: number;
+    constructor(sceneObjHolder:SceneObjHolder, infoIter:JMapInfoIter,parent: OtaKing, stepOffset: number){
+        super(sceneObjHolder, "OtaKingLongFoot", "OtaKingLongFoot", parent , DrawBufferType.EnemyDecoration);
+        this.stepOffset = stepOffset
```
magcius, line 5100: You can move the `private stepOffset: number` into the constructor for slightly easier init.

src/SuperMarioGalaxy/Actors/Enemy.ts
```diff
@@ -5053,3 +5054,223 @@ export class Petari extends LiveActor<PetariNrv> {
...
+        super.requestArchives(sceneObjHolder, infoIter);
+        sceneObjHolder.modelCache.requestObjectData('OtaKingLongFoot');
+    }
+
+}
+
+enum OtaKingNrv { Wait, ThrowCocoNut, ThrowFireBall }
+
+export class OtaKing extends LiveActor<OtaKingNrv> {
+    private feet: Array<PartsModel>;
```
magcius, line 5122: `private feet: PartsModel[]`

src/SuperMarioGalaxy/Actors/Enemy.ts
```diff
@@ -5053,3 +5054,223 @@ export class Petari extends LiveActor<PetariNrv> {
...
+    private longFeet: Array<PartsModel>;
+    private coconuts: Array<CocoNutBall>;
+    private currentMove: number;
+    private magma: PartsModel;
+    private magmaBloom: PartsModel;
+    private isLv2: boolean;
```
magcius, line 5128: This is never set.

src/SuperMarioGalaxy/Actors/Enemy.ts
```diff
@@ -5053,3 +5054,223 @@ export class Petari extends LiveActor<PetariNrv> {
...
+        //this.magma = createPartsModelEnemyAndFix(sceneObjHolder, this, 'OtaKingMagma',null,null,null,null);
+        this.magma = new PartsModel(sceneObjHolder, 'OtaKingMagma', 'OtaKingMagma', this, DrawBufferType.EnemyDecoration);
+        this.magmaBloom = new PartsModel(sceneObjHolder, 'OtaKingMagmaBloom', 'OtaKingMagmaBloom', this, DrawBufferType.BloomModel);
+        this.magmaBloom.initFixedPositionJoint(null, this.translation, null);
+        startBck(this.magma, 'wait');
+        startBck(this, 'wait');
+
+        this.initLongFoot(sceneObjHolder, infoIter);
+
+        for (let i in range(0,3)){
```
magcius, line 5256: `for (let i = 0; i < 3; i++) {`

src/SuperMarioGalaxy/Actors/MiscActor.ts
```diff
@@ -158,13 +158,13 @@ export class PartsModel extends LiveActor {
         connectToScene(sceneObjHolder, this, movementType, calcAnimType, drawBufferType, -1);
     }
 
-    public initFixedPositionMtxRelative(mtx: ReadonlyMat4, localTrans: ReadonlyVec3 | null): void {
-        this.fixedPosition = new FixedPosition(mtx, localTrans);
+    public initFixedPositionMtxRelative(mtx: ReadonlyMat4, localTrans: ReadonlyVec3 | null, localRot?: ReadonlyVec3 | null): void {
```
magcius, line 161: `localRot: ReadonlyVec3 | null = null` is what I typically use.

src/SuperMarioGalaxy/Actors/Enemy.ts
cbondurant: This is part of some code that I am still working on, that I maybe should have just left out completely but I worried that I would just be submitting more errors that way. The reason this doesnt throw in the current code is because the king does not launch any coconuts at this time, so they are always dead and this does not get called. 

I can take it out if needed.

src/SuperMarioGalaxy/Actors/Enemy.ts
```diff
@@ -5053,3 +5054,226 @@ export class Petari extends LiveActor<PetariNrv> {
...
+        vec3.sub(scratchVec3b, scratchVec3a, this.translation);
+        let tan = Math.atan2(scratchVec3b[0], scratchVec3b[2]);
+        let mod = ((MathConstants.TAU + (tan - (this.rotation[1]-MathConstants.TAU/2)))%MathConstants.TAU) + (this.rotation[1] - MathConstants.TAU/2);
+        let difference = mod - this.rotation[1];
+        if (((0 < difference) && (this.currentMove < difference))||((difference < 0) && (difference < this.currentMove))){
+            if (difference * this.currentMove < 0,0){
+                this.currentMove = 0.0;
+            }
+            this.rotation[1] += this.currentMove;
+            if (difference <= 0.0){
```
magcius, line 5165: I know Ghidra tends to invert conditionals, but I do try to lay out the conditionals to be a bit more how I would write it "naturally". I think this is all equivalent to:

```ts
const turnAmount = Math.sign(difference) * (0.03 * MathConstants.DEG_TO_RAD);
const turnAmountMax = MathConstants.DEG_TO_RAD;
this.currentMove = clampRange(this.currentMove + turnAmount, turnAmountMax);
```

I might also rename "currentMove" to "turnAmount"

src/SuperMarioGalaxy/Actors/Enemy.ts
```diff
@@ -5053,3 +5054,226 @@ export class Petari extends LiveActor<PetariNrv> {
...
+                this.currentMove += 0.003*MathConstants.DEG_TO_RAD;
+                if (MathConstants.DEG_TO_RAD <= this.currentMove){
+                    this.currentMove = MathConstants.DEG_TO_RAD;
+                }
+            }
+        }else{
+            this.rotation[1] = mod;
+            this.currentMove = 0;
+        }
+        this.rotation[1] = (this.rotation[1]+MathConstants.TAU)%MathConstants.TAU;
```
magcius, line 5181: You don't need this line.

src/SuperMarioGalaxy/Actors/Enemy.ts
```diff
@@ -5053,3 +5054,226 @@ export class Petari extends LiveActor<PetariNrv> {
...
+    }
+
+    private tryThrowCoconutOrFireBallIfWait(step: number): boolean {
+        if (isGreaterStep(this, step) || !this.isValidThrowCoconut()){
+            if (isGreaterStep(this, step << 1) || !this.isValidThrowFireball()){
```
magcius, line 5186: `step << 1` is equivalent to `step * 2.0`, please use that.

src/SuperMarioGalaxy/Actors/Enemy.ts
```diff
@@ -5053,3 +5054,226 @@ export class Petari extends LiveActor<PetariNrv> {
...
+        if (!this.isLv2){
+            let stepOffset = [3, 8, 0xE, 0x12];
+            for (let i = 0; i < 4; i++){
+                let longFoot = new OtaKingLongFoot(sceneObjHolder, infoIter, this, stepOffset[i]);
+                this.longFeet[i] = longFoot;
+                vec3.set(longFoot.scale, 1.0, 0.9, 0.8);
+            }
+
+            vec3.add(this.longFeet[0].translation, vec3.set(scratchVec3a, 735.0, 80.0, -55.0), this.translation);
+            vec3.set(this.longFeet[0].rotation, -9*MathConstants.DEG_TO_RAD, 266*MathConstants.DEG_TO_RAD, 0.0);
```
magcius, line 5225: Spaces between operands please.

src/SuperMarioGalaxy/Actors/Enemy.ts
```diff
@@ -5053,3 +5054,226 @@ export class Petari extends LiveActor<PetariNrv> {
...
+                startBck(this.feet[i], 'wait');
+            }
+        }
+
+        //this.magma = createPartsModelEnemyAndFix(sceneObjHolder, this, 'OtaKingMagma',null,null,null,null);
+        this.magma = new PartsModel(sceneObjHolder, 'OtaKingMagma', 'OtaKingMagma', this, DrawBufferType.EnemyDecoration);
+        this.magmaBloom = new PartsModel(sceneObjHolder, 'OtaKingMagmaBloom', 'OtaKingMagmaBloom', this, DrawBufferType.BloomModel);
+        this.magmaBloom.initFixedPositionJoint(null, this.translation, null);
+        startBck(this.magma, 'wait');
+        startBck(this, 'wait');
```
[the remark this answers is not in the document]
cbondurant: I would assume so, however I had actually forgotten about magmaBloom due to it seemingly not rendering in noclip. I've yet to investigate why this is happening.

src/SuperMarioGalaxy/Actors/Enemy.ts
```diff
@@ -5053,3 +5054,226 @@ export class Petari extends LiveActor<PetariNrv> {
...
+        }
+    }
+
+    private dirToPlayer(sceneObjHolder: SceneObjHolder){
+        getPlayerPos(scratchVec3a, sceneObjHolder);
+        vec3.sub(scratchVec3b, scratchVec3a, this.translation);
+        let tan = Math.atan2(scratchVec3b[0], scratchVec3b[2]);
+        let mod = ((MathConstants.TAU + (tan - (this.rotation[1]-MathConstants.TAU/2)))%MathConstants.TAU) + (this.rotation[1] - MathConstants.TAU/2);
+        let difference = mod - this.rotation[1];
+        if (((0 < difference) && (this.currentMove < difference))||((difference < 0) && (difference < this.currentMove))){
```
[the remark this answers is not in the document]
cbondurant: I dont think that quite works, because in the case of a negative distance, -1 for example, we want to know if `this.currentMove` overshoots distance.

so in the case of `distance = -1 && this.currentMove = -2` we should expect the conditional to be false because using the current move we would overshoot the goal distance, and instead should go to the else case were we just set it to the goal angle. However `abs(distance) > this.currentMove` would evaluate to true and would cause a false positive.

I dont think there is a way to rewrite this without causing errors in edge cases.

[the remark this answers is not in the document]
magcius: I meant `abs(distance) > abs(this.currentMove)`, which work with your example. The original expression is `(difference > 0 && difference > this.currentMove) || (difference < 0 && difference < this.currentMove)`; I would at least to refactor to that cleaned up expression.

cbondurant:
This PR is most likely going to be on hiatus, as I have started school again and do not have the time I did to work on it.

magcius:
That's fine. It'll still be here waiting if you ever decide to return.

PR #385, opened 2021-01-16
Add legacy rendering path for WebXR

fordacious:


src/gfx/platform/GfxPlatformWebGL2.ts
```diff
@@ -332,12 +334,14 @@ class GfxRenderPassP_GL implements GfxRenderPass {
...
 enum HostAccessPassCmd { uploadBufferData = 491, uploadTextureData, end };
+let g0 = 0;
 class GfxHostAccessPassP_GL implements GfxHostAccessPass {
     public u32: Growable<Uint32Array> = new Growable((n) => new Uint32Array(n));
     public gfxr: (GfxResource | null)[] = [];
     public bufr: ArrayBufferView[] = [];
 
-    public reset() { this.u32.r(); this.gfxr.length = 0; this.bufr.length = 0; }
+    public g0 = 0;
+    public reset() { this.g0 = ++g0; this.u32.r(); this.gfxr.length = 0; this.bufr.length = 0; }
```
magcius, line 344: ???

magcius (changes requested):
This is a bit overcomplicated. The issue is that GLES doesn't let us glBlitFramebuffer from an MSAA FB to another MSAA FB, and the WebXR framebuffer is MSAA. We need to round-trip to a texture to resolve MSAA FB -> single-sampled texture, and then paint the texture on the screen -- that's the "blitFullscreenTexture".

The way I would do this is the following: in configureSwapChain, if platformFramebuffer is passed in, assume it's a multi-sampled FB. In that case, create a texture there, like how you do in configureSwapChain_legacy, except you can just make it unified. It shouldn't be _scTexture, it should be named something like _scResolveTexture. Also set a bool, _scPlatformFramebufferIsMSAA.

In endPass, when we're resolving to the onscreen, when _scPlatformFramebufferIsMSAA, then we need to round-trip to the texture and fullscreen it back to the platform framebuffer. There shouldn't be any need to create "legacy" modes, we just need to handle this case correctly. In theory, it should also work without WebXR as well.

magcius:
Closing this -- I've since added an alternative renderer path for WebXR.

PR #393, opened 2021-03-07
ui: Fix Studio mode

VelocityRa:


src/ui.ts
```diff
@@ -1506,14 +1506,13 @@ class ViewerSettings extends Panel {
 
         this.contents.style.lineHeight = '36px';
 
+         // Don't change the order
```
VelocityRa, line 1509: bit lazy, lemme know if that's bad

magcius: I mean this has always needed to match with camControllerClasses.

PR #408, opened 2021-05-01
Give more descriptive name to teaser map

MysteryPancake:
This map is from the [Super 8 Interactive Teaser](https://combineoverwiki.net/wiki/Super_8_Interactive_Teaser) in the Extras menu.


magcius:
I'm good, thanks.

PR #412, opened 2021-05-25
Add names to Portal levels(HAS NOT BEEN BUILT)

stopmotio:
The test numbers ingame and the level name numbers are different which created some friction before
Never worked in typescript and made my work mimic mariokartwii code assuming that would work, let me know if i need to make other changes :)

PS I can't build this thing currently
Oh also there are some other errors from newer versions of the Source Engine renderer that I have no idea how to fix especially with expensive water

Thanks for all your work, you got so much working the last 6 months you make it look like child's play


stopmotio:
The reason I did these names was because testchmb_15 was actually 19, which I thought would be confusing if someone was looking for a particular chamber. It is you decision, but this is why I did this.

PR #421, opened 2021-07-28
Add Spongebob Squarepants: Revenge of the Flying Dutchman

Jellonator:
Hi! Over the past week I've been working on adding support for ROTFD to noclip.website. I'll also probably make a PR for Jimmy Neutron: Boy Genius in the future since they use the same engine and file formats. This implementation makes use of the Gamecube version's files, but the game was also released on the PS2.

### Data folder
In the folder extracted from an iso, copy the contents of the 'DATA' folder to the server's 'data/rotfd'. It should look like this:

```
data/
 |-rotfd/
   |-LEGAL.DGC
   |-PRELOAD.DGC
   |-BB/
   | |-LVL_BBEX.DGC
   | |-etc.
```

### Outside changes
The only changes I made outside of the `SpongebobROTFD` subfolder are for editing `main.ts`, adding RGBA8 support to the GX texture loader, and a function to convert `quat` to euler angles.

### Known issues
The main things that are missing as of now are particle effects and skeletal animations.

I still also haven't looked into the material formats close enough, so there's likely a few things that don't render 100% correctly. However, it's probably close enough that most people wouldn't notice unless they're looking close enough.

Also, the lighting is a bit darker than it is in-game, and I'm not sure why. I know that the game has a bloom filter, and maybe it also has some kind of color correction? It could also be that my lighting implementation is incorrect.

There's also a few issues that I'm having with transparency and layering. This is most evident in the Treedome areas.

Script objects (the red squares with yellow diamonds/crosses on them) are still visible, since they aren't made invisible by a material flag, unlike clip surfaces.


magcius:
Reviewing will take me a few days; please be patient and sorry about the wait.

> adding RGBA8 support to the GX texture loader

RGBA8 palettes don't exist on the GX hardware, so I'd prefer not to add this functionality to the GX subsystem directly. Do you know how the original engine handled this?

Jellonator:
> Reviewing will take me a few days; please be patient and sorry about the wait.
> 
> > adding RGBA8 support to the GX texture loader
> 
> RGBA8 palettes don't exist on the GX hardware, so I'd prefer not to add this functionality to the GX subsystem directly. Do you know how the original engine handled this?

No worries, take you time!

Unfortunately, I don't know how the original engine handles the RGBA8 palette. Neither ROTFD nor Jimmy Neutron contain debug symbols on Gamecube (although Jimmy does have them on PS2), so I've been a bit hesitant to decompile it. In the meantime, I can remove my changes to the GX subsystem, and figure out a different method of doing this instead.

magcius:
> Unfortunately, I don't know how the original engine handles the RGBA8 palette.

Dolphin has a setting somewhere which writes the texture format into the top left of the texture as an overlay. You might be able to use that to figure out how the game presents such textures to the hardware.

magcius (commented):
As a general overview, I let contributors define the code styles for their own game renderers; most comments are simply suggestions. However, there are a few points I'm stronger on, when code outside of the game renderer is changed.

One thing I've been trying to clean up is to just remove a lot of the more obtuse acronyms. I would prefer for the game data to be placed in a folder named "SpongeBobReturnOfTheFlyingDutchman", rather than the more inscrutable "rotfd"; I will forget what that stands like and think it's a PPC instruction or something. That applies both to the code directory name, the data name, and the URL ID. I know it's quite a bit more verbose, but I really would prefer it.

And yes, I know there still are plenty of games that don't follow this rule -- I really should clean the rest of them up where possible.

Jellonator:
I updated the PR.
I removed my changes to GX, and I completely removed support for the RGBA8 palette since it's only used in a single menu screen overlay anyways (which I didn't realize at the time lol).
I renamed the source and data folders to `src/SpongebobRevengeOfTheFlyingDutchman` and `data/SpongebobRevengeOfTheFlyingDutchman`.
I changed some of the constants to instead calculate based on the original string value, to make it a bit more clear what it's doing.
I also fixed a couple incorrect materials (namely the door on spongebob's house having the pineapple texture), and it now preloads the SPONGE.DGC file (e.g. for the pants/health pickup textures).

src/SpongebobRevengeOfTheFlyingDutchman/render.ts
```diff
@@ -510,7 +420,7 @@ class MeshRenderer {
         offs += fillVec4(d, offs, col.r, col.g, col.b, col.a);
         offs += fillVec4(d, offs, emit.r, emit.g, emit.b, 0);
         const tx = this.material.transform;
-        offs += fillMatrix4x4(d, offs, [
+        offs += fillMatrix4x2(d, offs, [
```
magcius, line 423: FYI, fillMatrix4x2 ignores the right-most two columns, so double-check you don't need access to the Z here.

src/SpongebobRevengeOfTheFlyingDutchman/program.glsl
```diff
@@ -0,0 +1,98 @@
...
+#endif
+
+#ifdef FRAG
+void main() {
+    vec4 texcol = texture(SAMPLER_2D(u_Texture), v_TexCoord);
+    vec3 surfacecol = texcol.rgb * u_Color.rgb;
+    float alpha = texcol.a * u_Color.a;
+    // SPECULAR
+    vec3 reflectLight = normalize(reflect(u_light.direction, v_ClipNormal));
+    vec4 reflectionColor = texture(SAMPLER_2D(u_TextureReflection), reflectLight.xy);
```
magcius, line 86: So you're flipping the world-space light direction over the view-space normal (it's not actually clip-space yet, because you don't multiply by the projection matrix), and using that as an input into a map?

That seems bizarre to me. If anything, I would expect both to be in world-space or both to be in view-space. But if that's what the game does, that's what the game does, I suppose. For the GameCube system, they would have had to do this in the GX material system, where the texture coordinates would be multiplied against a constant matrix computed on the CPU; there was no per-pixel texture math, only interpolated vertex texture coordinates.

Jellonator: Honestly, I don't know how the game is doing this. I kinda just put this in, thought it looked 'good enough for now,' and haven't messed with it too much since. Which, honestly, is kinda true for most of this shader. Do you recommend any resources on figuring this out, like GX documentation, or some kind of dolphin debugging tool, etc.?

Issue #424, opened 2021-08-06
Rewrite BCn texture utilities


magcius:
* [ ] Rewrite software BCn texture utilities in Rust (for performance)
* [ ] Make utilities sRGB aware
* [ ] Make utilities optionally automatically use queryTextureFormatSupported
* [ ] Expand queryTextureFormatSupported to include w/h queries to see if we can use native hardware decoding

Jellonator:
https://github.com/Jellonator/noclip.website/tree/bcn
Finished first task. Moved deswizzle and BCn decompression to Rust. I managed to eek out a bit more performance by performing both in the same function.
My (admittedly not exactly rigorous) testing seems to show that it is about 40% faster than before.

Jellonator:
I made the texture loading bit async, dunno if you want to keep that or not. It's difficult to make it synchronous due to the way loading WASM works.

magcius:
Can you file a PR so I can review it?

magcius:
Expanding on the notes for the other checkboxes.

* [ ] sRGB-aware blending only needs to happen when computing the blended midpoints for the color tables. The way I expect it's going to work, we'll decode sRGB BC1 textures to an RGBA_U8_SRGB texture, so the endpoints should already be correct. We won't be able to get blending 100% correct for magnified textures, but we can do our best for the midpoints.

* [ ] queryTextureFormatSupported needs to accept width and height as arguments, and only return true for the BCn formats if they are divisible by 4 on the WebGL 2 backend. WebGPU could potentially support non-block-aligned mip textures as an extension, but should also do the same check for now, see https://github.com/gpuweb/gpuweb/issues/2006
* [ ] Non-swizzling decoding could should also be added to Rust. Maybe there's some clever way to support both kinds of deswizzlers using a polymorphic struct so we can keep the combined deswizzle/pixel decode system (see something similar in gx_texture.rs)
* [ ] The high-level game-facing API for this that's going to replace bc_texture.ts should be adapted to ask the device backend if it supports native hardware textures, and if so, uses that directly, and if it doesn't, falls back to decoding on the HW.

For game support, currently:
* [ ] Source Engine and The Witness assume hardware BC1 support and does not have software decoding as a fallback
* [ ] Psychonauts always decompresses
* [ ] GTA has the option of using Basis textures, but its TXD fallback supports both hardware and non-hardware BC1
* [ ] Dark Souls supports supports both hardware and non-hardware BC1

[...]

PR #445, opened 2021-08-11
Rewrite BCn texture utilities in Rust (+ deswizzling)

Jellonator:
PR for first task of #424


rust/src/util.rs
```diff
@@ -0,0 +1,115 @@
...
+    unsafe { *mem::transmute::<&u8, &u32>(&src[offs]) & 0x00FFFFFF }
+}
+
+#[cfg(target_endian="big")]
+pub fn get_uint32_be(src: &[u8], offs: usize) -> u32 {
+    unsafe { *mem::transmute::<&u8, &u32>(&src[offs]) }
+}
+
+#[cfg(target_endian="little")]
+pub fn get_uint16_be(src: &[u8], offs: usize) -> u16 {
```
magcius, line 103: This implementation should work in both configurations. I don't know why we'd ever be on a big-endian target on a website in 2021, and since this version works for both, I'd prefer to just keep this version for everything.

rust/src/util.rs
```diff
@@ -0,0 +1,115 @@
...
+
+pub fn halfblend(a: u8, b: u8) -> u8 {
+    let a = a as u32;
+    let b = b as u32;
+    let ret = (a + b) >> 1;
+    ret as u8
+}
+
+// Use the fast GX approximation.
+pub fn s3tcblend(a: u8, b: u8) -> u8 {
```
magcius, line 43: I don't see a value in sharing this. tbh we should probably use a better approximation for non-GX textures, since what we really want is 1/3 and 2/3 rather than 3/8 and 5/8 (or a fixed-point version that's good enough for the U8 range).

rust/src/bcn_texture.rs
```diff
@@ -0,0 +1,706 @@
+use wasm_bindgen::prelude::*;
+use crate::util;
+
+const GOB_SIZE_X: usize = 64;
+const GOB_SIZE_Y: usize = 8;
+
+#[wasm_bindgen]
+#[derive(Copy, Clone, PartialEq, Eq, Debug)]
```
magcius, line 8: We don't need all these derives, do we?

rust/src/bcn_texture.rs
```diff
@@ -0,0 +1,706 @@
+use wasm_bindgen::prelude::*;
```
magcius, line 1: I do not like * imports, please don't use them.

rust/src/bcn_texture.rs
```diff
@@ -0,0 +1,706 @@
...
+use crate::util;
+
+const GOB_SIZE_X: usize = 64;
+const GOB_SIZE_Y: usize = 8;
+
+#[wasm_bindgen]
+#[derive(Copy, Clone, PartialEq, Eq, Debug)]
+pub enum ChannelFormat {
```
magcius, line 9: This is really not BCn texture, it's Tegra deswizzling and Tegra decoding. Please rename this to tegra_texture.rs.

rust/src/bcn_texture.rs
```diff
@@ -0,0 +1,706 @@
...
+    }
+    dst
+}
+
+fn color_table_bc1(color_table: &mut [u8; 16], color1: u16, color2: u16) {
+    // Fill in first two colors in color table.
+    // TODO(jstpierre): SRGB-correct blending.
+    color_table[0] = util::expand5to8(((color1 >> 11) & 0x1F) as u8);
+    color_table[1] = util::expand6to8(((color1 >> 5) & 0x3F) as u8);
+    color_table[2] = util::expand5to8((color1 & 0x1F) as u8);
```
magcius, line 166: I know you moved the color_table[3] = 0xFF; out here for a slight micro-optimization, but unless it makes a significant difference, I would leave it as before.

Note that this representation of a color_table of 16 u8's was optimized for speed in a GC-sensitive tight loop. It's possible that with Rust's compiler and several additional features, we could make this code a lot more ergonomic, first by possibly having a color_table be a length-4 array of four U8's, or an array of U32's instead.

That's not blocking the review though.

src/fres_nx/tegra_texture.ts
```diff
@@ -269,3 +269,80 @@ export function translateImageFormat(imageFormat: ImageFormat): GfxFormat {
...
+    const depth = textureEntry.depth;
+    const channelFormat = getChannelFormat(textureEntry.imageFormat);
+    const typeFormat = getTypeFormat(textureEntry.imageFormat);
+    const info = {
+        width,
+        height,
+        depth,
+    }
+    switch (channelFormat) {
+    case ChannelFormat.Bc1:
```
magcius, line 286: Please use switch statements in JS/TS as infrequently as possible.

rust/src/gx_texture.rs
```diff
@@ -1,56 +1,26 @@
...
-}
+use crate::util;
 
 fn decode_rgb5a3_to_rgba8(dst: &mut[u8], p: u16) {
     if (p & 0x8000) != 0 {
         // RGB5
-        dst[0] = expand_n_to_8(5, ((p >> 10) & 0x1F) as u8);
-        dst[1] = expand_n_to_8(5, ((p >>  5) & 0x1F) as u8);
-        dst[2] = expand_n_to_8(5, ((p >>  0) & 0x1F) as u8);
+        dst[0] = util::expand5to8(((p >> 10) & 0x1F) as u8);
```
magcius, line 7: My hope is that expand_n_to_8 would be inlined / constant folded -- I don't see the reason to re-introduce the older expandNto8 functions if we have the generic form.

rust/src/util.rs
```diff
@@ -0,0 +1,115 @@
...
+
+pub fn expand5to8(n: u8) -> u8 {
+    (n << (8 - 5)) | (n >> (10 - 8))
+}
+
+pub fn expand6to8(n: u8) -> u8 {
+    (n << (8 - 6)) | (n >> (12 - 8))
+}
+
+pub fn halfblend(a: u8, b: u8) -> u8 {
```
magcius, line 35: This is not used outside of CMPR. Either we should make the new system use it, or we have no reason to keep it a separate function. I think we should just make CMPR adopt the same expression you added in the generic BC decoders with the cast to u16.

src/fres_nx/tegra_texture.ts
```diff
@@ -269,3 +269,80 @@ export function translateImageFormat(imageFormat: ImageFormat): GfxFormat {
         throw "whoops";
     }
 }
+
+export async function decompress_bcn_deswizzle(textureEntry: BRTI, pixels: Uint8Array): Promise<DecodedSurfaceSW> {
```
magcius, line 273: Please use `decompress`, the underscore naming scheme is not really that prevalent in the TypeScript code style.

src/fres_nx/tegra_texture.ts
```diff
@@ -269,3 +269,80 @@ export function translateImageFormat(imageFormat: ImageFormat): GfxFormat {
...
+    let wasm = await import("../../rust/pkg/index");
```
magcius, line 274: Use `const` please.

src/fres_nx/tegra_texture.ts
```diff
@@ -269,3 +269,80 @@ export function translateImageFormat(imageFormat: ImageFormat): GfxFormat {
...
+                    channelFormat === ChannelFormat.Bc4 ? wasm.BCNType.BC4 : wasm.BCNType.BC5,
+                    wasm.SurfaceFlag.UNorm,
+                    width,
+                    height,
+                    depth,
+                    pixels,
+                    textureEntry.blockHeightLog2
+                ),
+            };
+        }
```
magcius, line 324: Put `} else {` on one line, please.

rust/src/lib.rs
```diff
@@ -7,6 +7,9 @@ static ALLOC: wee_alloc::WeeAlloc = wee_alloc::WeeAlloc::INIT;
...
 pub mod yaz0;
 pub mod gx_texture;
+pub mod bcn_texture;
+pub mod util;
 
-pub use crate::yaz0::yaz0dec;
-pub use crate::gx_texture::decode_texture;
+pub use crate::yaz0::*;
+pub use crate::gx_texture::*;
+pub use crate::bcn_texture::*;
```
magcius, line 15: Please do not use * imports.

magcius (changes requested):

rust/src/gx_texture.rs
Jellonator: I wasn't sure about this, so I checked the disassembly in Godbolt. You're right, it does get optimized/inlined to the same instructions. I'll keep it as expand_n_to_8 in that case.

rust/src/util.rs
Jellonator: Should the CMPR format use the same 1:2/2:1 color ratio as Tegra?

magcius: The implementation of halfblend in CMPR is identical to where it's similarly used in the BC1 decoder, it's just extracted to a function in one of them, but not the other.

src/fres_nx/tegra_texture.ts
Jellonator: The 'decompress' function is also used by Mario Odyssey. I'm gonna test and make sure SMO is fine with this change too.

rust/src/util.rs
Jellonator: I'll remove the versions of the functions that target big endian, but I think it may be best to keep the cfg attributes just in case. I did a quick search before implementing this, and saw that there might be plans for WASM to support big endian devices. Take this with a grain of salt though because I admittedly did not look into it that hard, I might have misread it, or it might have been some kind of translation layer.

magcius: The implementation `((src[offs] as u16) << 8) | (src[offs+1] as u16)` should work on both little-endian and big-endian systems.

Jellonator: oh nevermind, I see what you mean. I'll change it to use the non-platform dependent functions then

Jellonator:
Implemented the sRGB blending as well, based on the formula I found here: https://entropymine.com/imageworsener/srgbformula/

magcius:
If you want to double-check, official reference for SRGB can be found here:

https://microsoft.github.io/DirectX-Specs/d3d/archive/D3D11_3_FunctionalSpec.htm#3.2.3.7%20SRGB%20-%3E%20FLOAT
https://microsoft.github.io/DirectX-Specs/d3d/archive/D3D11_3_FunctionalSpec.htm#3.2.3.8%20FLOAT%20-%3E%20SRGB

src/fres_nx/tegra_texture.ts
Jellonator: SMO textures seem to decompress fine

rust/src/tegra_texture.rs
```diff
@@ -0,0 +1,493 @@
...
+#[derive(Clone)]
+pub struct SurfaceMetaData {
+    width: usize,
+    height: usize,
+    depth: usize,
+    block_height_log2: usize,
+}
+
+pub trait OutputType
+{
```
magcius, line 106: I think the proper code style here is same-line braces.

rust/src/util.rs
```diff
@@ -0,0 +1,95 @@
...
+        (value + 0.5) as u8
+    }
+}
+
+pub type BlendFunction<T, U> = fn(T, T, U, U) -> T;
+
+pub fn blend_srgb_u8(a: u8, b: u8, weight_a: u32, weight_b: u32) -> u8 {
+    let a = srgb_to_linear(a);
+    let b = srgb_to_linear(b);
+    let v = (a * (weight_a as f32) + b * (weight_b as f32)) / ((weight_a as f32) + (weight_b as f32));
```
magcius, line 61: Any reason you're choosing to pass weights as u32's instead of f32's? Is it to prepare for a pure integer version in case we want to avoid float math in the future? If not, I'd prefer to just have this be used like: `blend_srgb_u8(a, b, 1.0f/3.0f, 2.0f/3.0f)`.

rust/src/tegra_texture.rs
```diff
@@ -0,0 +1,493 @@
...
+    fn read(value: u8) -> Self::Type;
+    fn narrow(value: Self::BigType) -> Self::Type;
+}
+
+pub struct OutputSigned {}
+impl OutputType for OutputSigned {
+    type Type = i8;
+    type BigType = i32;
+    const MIN: i8 = -128;
+    const MAX: i8 = 127;
```
magcius, line 120: Kinda surprised Rust doesn't have a std::numerical_limits<i8> kinda thing built-in?

rust/src/gx_texture.rs
```diff
@@ -1,56 +1,42 @@
...
-        _ => unreachable!(),
-    }
-}
-
-fn get_u16_be(src: &[u8], i: usize) -> u16 {
-    u16::from_be_bytes(src[i..i+2].try_into().unwrap())
-}
-
-fn s3tcblend(a_: u8, b_: u8) -> u8 {
+// Use the fast GX approximation.
```
magcius, line 4: This comment doesn't quite make as much sense to me, when we're already in the GX implementation.

rust/src/gx_texture.rs
```diff
@@ -1,56 +1,42 @@
...
+pub fn s3tcblend(a: u8, b: u8) -> u8 {
     // return (a*3 + b*5) / 8;
-    let a = a_ as u32;
-    let b = b_ as u32;
-    let tmp = (((a << 1) + a) + ((b << 2) + b)) >> 3;
-    return tmp as u8;
+    let a = a as u32;
```
magcius, line 7: I'm surprised `let a = a` compiles; I figured Rust would warn on shadowing. I kinda don't like shadowing though, so I would prefer if you changed it back to `a_`.

rust/src/gx_texture.rs
```diff
@@ -1,56 +1,42 @@
...
-    }
-}
-
-fn get_u16_be(src: &[u8], i: usize) -> u16 {
-    u16::from_be_bytes(src[i..i+2].try_into().unwrap())
-}
-
-fn s3tcblend(a_: u8, b_: u8) -> u8 {
+// Use the fast GX approximation.
+pub fn s3tcblend(a: u8, b: u8) -> u8 {
```
magcius, line 5: No reason for this (and halfblend) to be pub anymore.

magcius (commented):
This is looking quite good! A few minor style nits, then this will be good to go.

PR #479, opened 2021-12-20
Add Wii Banners

ddrmax34:
I want support for Wii banners, could you actually implement them?


magcius:
I have no clue what you did, but this PR doesn't include any Wii banner support as far as I can tell...

PR #480, opened 2021-12-21
Wii banner support

ddrmax34:
Could you update the repository so you can implement banner support?


magcius:
Still no Wii banner support

PR #497, opened 2022-02-17
Add Unity asset reader + Subnautica scene

wgreenberg:
So far it's just a barebones render of the scanner map, but it's a start!


src/Subnautica/scenes.ts
```diff
@@ -0,0 +1,241 @@
...
+}
+
+const bindingLayouts: GfxBindingLayoutDescriptor[] = [
+    { numUniformBuffers: 1, numSamplers: 0 }, // ub_SceneParams
+];
+
+class Mesh {
+    vertices: Float32Array,
+    normals: Float32Array,
+    indices: Int32Array,
```
magcius, line 107: This should probably be an `interface`, instead of a `class`, and please use semicolons instead of commas.

src/Subnautica/scenes.ts
```diff
@@ -0,0 +1,241 @@
...
+
+    public destroy(device: GfxDevice) {
+        device.destroyInputLayout(this.inputLayout);
+        device.destroyProgram(this.program);
+        this.meshRenderers.forEach((r) => r.destroy(device));
+        this.renderHelper.destroy();
+    }
+}
+
+class Vertex {
```
magcius, line 193: Mind using `vec3` from gl-matrix instead?

magcius, line 193: Mind using `vec3` from gl-matrix instead?

src/Subnautica/scenes.ts
```diff
@@ -0,0 +1,241 @@
...
+    public trisBuf: GfxBuffer;
+    public inputState: GfxInputState;
+    public numVertices: number;
+
+    constructor(device: GfxDevice, mesh: Mesh, public offset: Vertex, public inputLayout: GfxInputLayout) {
+        for (let i=0; i<mesh.vertices.length; i += 3) {
+            mesh.vertices[i+0] += offset.x;
+            mesh.vertices[i+1] += offset.y;
+            mesh.vertices[i+2] += offset.z;
+        }
```
magcius, line 72: Could you move this offset into a model matrix UBO? Up to you, but I think that would be helpful going forward.

src/Subnautica/scenes.ts
```diff
@@ -0,0 +1,241 @@
...
+
+        this.inputState = device.createInputState(inputLayout, [
+            { buffer: this.vertsBuf, byteOffset: 0, },
+            { buffer: this.normsBuf, byteOffset: 0, },
+        ], { buffer: this.trisBuf, byteOffset: 0 });
+        this.numVertices = mesh.indices.length;
+    }
+
+    public prepareToRender(renderInstManager: GfxRenderInstManager): void {
+        const renderInst = renderInstManager.newRenderInst();
```
magcius, line 86: Consider frustum culling on the AABB bounds?

src/viewer.ts
```diff
@@ -378,7 +378,7 @@ async function initializeViewerWebGL2(out: ViewerOut, canvas: HTMLCanvasElement)
         return InitErrorCode.GARBAGE_WEBGL2_SWIFTSHADER;
 
     const config = new GfxPlatformWebGL2Config();
-    config.trackResources = IS_DEVELOPMENT;
+    config.trackResources = false;
```
magcius, line 381: Debug change

src/Common/Unity/AssetManager.ts
```diff
@@ -0,0 +1,88 @@
...
+function concatBufs(a: Uint8Array, b: Uint8Array): Uint8Array {
+    let result = new Uint8Array(a.byteLength + b.byteLength);
+    result.set(a);
+    result.set(b, a.byteLength);
+    return result;
+}
+
+interface Range {
+    rangeStart: number,
+    rangeSize: number,
```
magcius, line 26: Please use semicolons, not commas

magcius (changes requested):

rust/Cargo.toml
```diff
@@ -7,15 +7,17 @@ edition = "2018"
 [lib]
 crate-type = ["cdylib", "rlib"]
 
+[package.metadata.wasm-pack.profile.release]
+wasm-opt = ['-g']
+
 [profile.release]
 lto = true
+debug = true
```
magcius, line 15: This change / the -g looks wrong?

rust/Cargo.toml
```diff
@@ -7,15 +7,17 @@ edition = "2018"
...
 
-[features]
-default = ["wee_alloc"]
-
 [dependencies]
 wasm-bindgen = "0.2.45"
 wee_alloc = { version = "0.4.5", optional = true }
 naga = { git = "https://github.com/gfx-rs/naga", features = ["glsl-in", "wgsl-out"] }
 web-sys = { version = "0.3.48", features = ["console"] }
+byteorder = "1.4.3"
```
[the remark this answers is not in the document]
wgreenberg: yeah, it's how i deserialize all the other types -- eventually i could probably replace it w/ the BE/LE functions you've got in util though

src/Subnautica/scenes.ts
wgreenberg: honestly not sure how to do this -- would you say it's a must-have, or something that could be implemented later?

src/Subnautica/scenes.ts
```diff
@@ -0,0 +1,211 @@
...
+}
+
+const id = 'Subnautica';
+const name = 'Subnautica';
+
+const sceneDescs = [
+    new SubnauticaSceneDesc("Scanner Map", "Scanner Map"),
+];
+
+export const sceneGroup: Viewer.SceneGroup = { id, name, sceneDescs };
```
magcius, line 211: Mind adding a `hidden: true` to here? That hides it from the main site, but still makes it visible when in development mode. Means I can merge this PR without having the data fully live yet.

src/Common/Unity/AssetManager.ts
```diff
@@ -0,0 +1,223 @@
...
+    let indices = mesh.get_index_data();
+    let indexBufferFormat: GfxFormat;
+    let numIndices = 0;
+    if (mesh.index_format === _wasm.IndexFormat.UInt32) {
+        indexBufferFormat = GfxFormat.U32_R;
+        numIndices = indices.length / 4;
+    } else {
+        indexBufferFormat = GfxFormat.U16_R;
+        numIndices = indices.length / 2;
+    };
```
magcius, line 216: Extra semicolon

Also, not a requirement to land, just a reminder that we should probably expose submeshes too. I need to do some investigation to figure out how submeshes map to materials, too.

src/Common/Unity/AssetManager.ts
```diff
@@ -0,0 +1,223 @@
...
+import { SceneContext } from '../../SceneBase';
+import { downloadBlob } from '../../DownloadUtils';
+import { AssetInfo, Mesh, AABB as UnityAABB, VertexFormat, StreamingInfo, ChannelInfo } from '../../../rust/pkg/index';
+import { GfxDevice, GfxBufferUsage, GfxInputState, GfxFormat, GfxInputLayout, GfxVertexBufferFrequency, GfxVertexAttributeDescriptor, GfxInputLayoutBufferDescriptor } from '../../gfx/platform/GfxPlatform';
+import { FormatCompFlags } from '../../gfx/platform/GfxPlatformFormat';
+import { assert } from '../../util';
+import * as Geometry from '../../Geometry';
+import { vec3 } from 'gl-matrix';
+
+let _wasm: any | null = null;
```
magcius, line 11: It would be nice if this wasn't marked as any. You should be able to use an import type now:

```ts
let _wasm: typeof import('../../../rust/pkg/index') | null = null;
```

src/Common/Unity/AssetManager.ts
```diff
@@ -0,0 +1,223 @@
...
+    let vertsBuf = makeStaticDataBuffer(device, GfxBufferUsage.Vertex, vertices.buffer);
+    let normsBuf = makeStaticDataBuffer(device, GfxBufferUsage.Vertex, normals.buffer);
+    let trisBuf = makeStaticDataBuffer(device, GfxBufferUsage.Index, indices.buffer);
+
+    let state = device.createInputState(layout, [
+        { buffer: vertsBuf, byteOffset: 0, },
+        { buffer: normsBuf, byteOffset: 0, },
+    ], { buffer: trisBuf, byteOffset: 0 });
+
+    return new UnityMesh(layout, state, indices.length, mesh.local_aabb);
```
magcius, line 159: Reminder that you're leaking the buffers here. If you want to turn on the leak checker, the easiest way to do that is to adjust `loadSceneDelta` from 1 to 0 in `main.ts`.

Maybe I should do that by default in development builds....

src/Common/Unity/AssetManager.ts
```diff
@@ -0,0 +1,223 @@
...
+        numIndices = indices.length / 4;
+    } else {
+        indexBufferFormat = GfxFormat.U16_R;
+        numIndices = indices.length / 2;
+    };
+
+    let layout = device.createInputLayout({ vertexAttributeDescriptors, vertexBufferDescriptors, indexBufferFormat });
+    let state = device.createInputState(layout, [
+        { buffer: makeStaticDataBuffer(device, GfxBufferUsage.Vertex, mesh.get_vertex_data()), byteOffset: 0 },
+    ], { buffer: makeStaticDataBuffer(device, GfxBufferUsage.Index, indices), byteOffset: 0 });
```
magcius, line 221: Again, reminder that you're leaking the buffers here.

rust/src/unity/mesh.rs
```diff
@@ -0,0 +1,585 @@
...
+    baked_convex_collision_mesh: Vec<u8>,
+    baked_triangle_collision_mesh: Vec<u8>,
+    streaming_info: StreamingInfo,
+}
+
+#[wasm_bindgen]
+impl Mesh {
+    pub fn from_bytes(data: Vec<u8>, asset: &AssetInfo) -> std::result::Result<Mesh, String> {
+        let mut reader = AssetReader::new(data);
+        reader.set_endianness(asset.header.endianness);
```
magcius, line 424: There's a part of me that believes we shouldn't see anything but little-endian in TYOOL2022 -- and I believe the new uncompressed mesh loading code relies on that. PowerPC is dead :pray:, we can give it up now. So we could probably turn this into an error / assert.

rust/src/unity/mesh.rs
```diff
@@ -0,0 +1,585 @@
...
+    pub fn is_compressed(&self) -> bool {
+        return self.mesh_compression != MeshCompression::Off
+    }
+
+    pub fn get_vertices(&self) -> Option<Vec<f32>> {
+        if self.mesh_compression == MeshCompression::Off {
+            None
+        } else {
+            Some(self.compressed_mesh.vertices.unpack())
+        }
```
magcius, line 441: Optional feedback, not required for merge:

I think it might be unclear that this is doing work to uncompress mesh data. I wonder if we should have an "unpack_mesh_data(channel_mask)" method that unpacks to a sequence of streams/channels, and then the TS code doesn't have to care about whether the data is compressed or not, it can just iterate over the final stream/channel setup.

For a compressed mesh, we'd have a stream per channel (e.g. a stream for positions, a stream for UVs, a stream for normals, etc.), while we'd continue to use the packed data on the Unity side.

rust/Cargo.toml
```diff
@@ -7,15 +7,16 @@ edition = "2018"
 [lib]
 crate-type = ["cdylib", "rlib"]
 
+[package.metadata.wasm-pack.profile.release]
+wasm-opt = ['-Os', '--debuginfo']
```
magcius, line 11: I'm still a bit scared about putting this into the final build; is there a way to only put this in webpack.dev.js mode?

src/Common/Unity/AssetManager.ts
```diff
@@ -0,0 +1,223 @@
...
+import { FormatCompFlags } from '../../gfx/platform/GfxPlatformFormat';
+import { assert } from '../../util';
+import * as Geometry from '../../Geometry';
+import { vec3 } from 'gl-matrix';
+
+let _wasm: any | null = null;
+
+async function loadWasm() {
+    if (_wasm === null) {
+        _wasm = await import('../../../rust/pkg/index');
```
magcius, line 15: It would also be great if the Rust package had namespaces, since just having a type called "AssetInfo" in the Rust code is a bit confusing, and if we do textures, we might run into conflicts with the GX texture loader code, but I think that's a taller order...

luv u 2 wasm-bindgen

magcius (commented):
This is so cool! Finally got it running locally, and it's been fun to mouse around. General note on review:

* As always, authors are given final say on code structure, API designs, and full artistic control for their contributed games, whatever that means ;). Please treat all my nitpicks and suggestions as *optional*, feel free to ignore them, and I will merge this PR without addressing any of them.

* Changes to common/shared code I might hold to a bit higher of a gatekeeping standard. There are two of those changes, really: the change to Geometry.ts (LGTM, harmless), and the Rust Cargo build changes, which I would like to understand a bit more carefully before merging, since it will affect existing games.

* Leak fixes are something that are not required for this PR to be merged, but are required before we make the game public on the site.

src/Common/Unity/AssetManager.ts
```diff
@@ -0,0 +1,201 @@
...
+}
+
+export interface MeshMetadata {
+    name: string;
+    offset: number;
+    size: number;
+}
+
+export const a_Position = 0;
+export const a_Normal = 1;
```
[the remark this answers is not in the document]
wgreenberg: ohh channel indices make way more sense!

rust/src/unity/mesh.rs
wgreenberg: funny enough, all unity asset files start out as big endian, and then switch endianness based on a value read in the header

magcius: I can imagine a big-endian header, but I would hope most of the actual these days past the header is little endian.

rust/src/unity/mesh.rs
wgreenberg: this sounds good to me, though i don't have a good idea of how to implement that yet. how about for now, i rename these to `unpack_*`?

magcius: As I mentioned, all suggestions are 100% optional. Feel free to rename, or simply not do anything!

rust/Cargo.toml
wgreenberg: stuck this in a dev-build-only config!

src/Common/Unity/AssetManager.ts
wgreenberg: fix'd

src/Common/Unity/AssetManager.ts
wgreenberg: yeahhh, maybe in future work i'll just rename everything `Unity*`

PR #503, opened 2022-03-19
Studio: Multi-Track Timeline & Major UI Overhaul

Veegie:
This PR features (another) major overhaul for Studio mode, with a new and improved UI (full-width, anchored to the bottom of the window), multi-track editing, manual keyframe value editing and editable interpolation types. For now, still hidden behind a console call to `main.ui.studioSidePanel.v()`.

Changes to common code, and other bits that are actually of note and may be worth a review/revision:

- All input handling and preview line drawing has been moved from `Camera.ts` into `Studio.ts`.
- The `options` member of `UI.RadioButtons` was made public.
- Added credits for new Zoom In/Zoom Out icons.
- The Keyframe interface has been modified to add a few new fields: Interpolation In Type, Interpolation Out Type, Ease In Coefficient, and Ease Out Coefficient. The coefficients have a range of [0-1] and adjust the magnitude of the auto-generated tangent values. Logic for loading animations has been updated to conform keyframes that are missing these fields. By default, the interpolation types are set to `InterpolationType.Ease` and the coefficients are set to 1.
- Studio settings (Show animation preview line/live preview/Auto-save) are now saved to local storage on modification.
- Re-added logic for normalizing bank rotation. The playback engine is still kept stateless; we simply get the normalized bank rotation value each frame.

Currently writing up some brief reference documentation, which I plan to add as a Wiki page on this repo.


magcius:
This all looks good to me.

I do have one possible UX note: could you save whether the user has enabled Studio to a preference so they don't need to type .v() at every page load? The initial toggle-on should still be from the console, but once a user enables it once, it should stay enabled.

Veegie:
> I do have one possible UX note: could you save whether the user has enabled Studio to a preference so they don't need to type .v() at every page load?

Sure, will do. The hope was to un-hide Studio Mode by default after doing some beta testing with this release and fixing any bugs that crop up. With your permission, of course.

PR #516, opened 2022-04-25
Names and Categories for Half-Life and Mario Odyssey

slambo311:
I didn't do ALL of the maps and categories (Half-Life maps aren't organized very well, and sometimes I just don't remember the actual names of some Mario Odyssey maps), but I did do a little bit.


slambo311:
I did another update for it, fixed quite a bit.

PR #521, opened 2022-05-04
Add mario odyssey names, yet again.

slambo311:
Finally figured out what I did wrong. One comma, ONE comma just broke it.


magcius:
One last chance. If this doesn't work, I won't merge any future PRs from you unless you're able to test them yourself.

PR #525, opened 2022-05-23
Even MORE Mario Odyssey naming

slambo311:
I checked it for errors around 20 times manually, but I didn't deploy it, since it's incredibly annoying to do with only a Pixel 2 phone and an outdated school Chromebook that can't run Linux stuff, so there still might be some errors.


magcius:
there are indeed errors

slambo311:
Alright.

PR #526, opened 2022-06-18
Super monkey ball 1: file drop, fixes, naomilib

ComplexPlane:
* Add GMA+TPL drag and drop support
* Fix issue where "null2" half ball model is shown on many stages
* Some NaomiLib (SMB Arcade Naomi model format) support. It's pretty busted right now but it's not used atm unless a Naomi model container+TPL is drag and dropped.


src/SuperMonkeyBall/AnimGroup.ts
```diff
@@ -0,0 +1,280 @@
+import { mat4, vec3 } from "gl-matrix";
```
ComplexPlane, line 1: Just a file split

src/SuperMonkeyBall/Lighting.ts
```diff
@@ -0,0 +1,54 @@
+import { vec3 } from "gl-matrix";
```
ComplexPlane, line 1: Just a file split

src/SuperMonkeyBall/World.ts
```diff
@@ -384,7 +97,75 @@ export class World {
         this.background.prepareToRender(ctx, this.lighting);
     }
 
+    public gettextureCache(): TextureCache {
```
magcius, line 100: bad capitalization here

PR #540, opened 2022-09-26
Here we go again, more SMO naming.

slambo311:
I checked it multiple times to see if I made any mistakes, didn't find any.


slambo311:
Most of the names are just the internal name, but with spaces and whatnot, but hey, it's better than JUST the internal name.

slambo311:
This has spent almost a third of a year without any comments or being closed, might as well close it now.

PR #558, opened 2022-12-16
Update Scenes_PokemonHGSS.ts

rjd1922:
As discussed on Discord. Hopefully the accented e won't break anything.


PFedak:
The new names/reordering looks great! I think for search reasons we'd better stick with regular e. 
Also, only if you want, another thing you could do is add more lines to that list which are just strings (like https://github.com/magcius/noclip.website/blob/master/src/SuperMarioGalaxy/Scenes_SuperMarioGalaxy1.ts#L162-L167 )
which will show up as headings, to organize things.

Happy to merge it once you make the "e" change

rjd1922:
I changed the accented e's back.

PR #571, opened 2023-02-02
Fix GoldSrc lightmap coordinates

byplt:
There is lightmap coordinate problem on goldsrc. As far as I investigate that the problem is caused from the minTexCoordS. This pull will fix majority problem but there is minor problem still exists, you can see compared images below. Mainly compared from (https://github.com/skyrim/hlviewer.js) , but it has same problem. There is a project is working perfectly, you can check out: (https://github.com/UnrealKaraulov/newbspguy)

compare

### DIFF
--- a/src/GoldSrc/BSPFile.ts
+++ b/src/GoldSrc/BSPFile.ts
@@ -237,8 +237,8 @@ export class BSPFile {
                 vertexData[dstOffsVertex++] = 0;
                 vertexData[dstOffsVertex++] = 0;
 
-                minTexCoordS = Math.min(minTexCoordS, texCoordS);
-                minTexCoordT = Math.min(minTexCoordT, texCoordT);
+                minTexCoordS = Math.min(minTexCoordS, Math.round(texCoordS));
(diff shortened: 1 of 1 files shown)


magcius:
Thank you! I pushed a simpler patch which should fix both artifacts you're seeing.

byplt:
The patch is not properly work, especially, you can see the left wall. I think you are missing something from ligthmap coordinates, lightmappackerpage or finding the min,max values. If you fix completely, I'm glad. Thanks. 
compare

magcius:
What map is that?

byplt:
boot_camp

magcius:
f62012ecbacff6c21b9b3f6bd72336ab3b43414c should fix that wall as well.

byplt:
Thanks, it works but as i mentioned that there are minor problem still exists. Unfortunately this minor problem becomes major problem some map. If you want to see, i can send screenshots. But, i think that this patch is good enough. 
compare

magcius:
f52a647b6cb8f7d1249f2227b2f71ea52021580a fixes this issue.

byplt:
Working perfectly, thanks.

Issue #575, opened 2023-02-24
Portal 2 Workshop breaks the mouse somehow


slambo311:
So after opening the workshop, just moving my mouse onto anything on the selection menu makes it load, despite me not clicking. Here's a poorly-made video:
[Screen recording 2023-02-24 10.56.44 AM.webm](https://user-images.githubusercontent.com/61334045/221225623-f7f669ac-8c9c-4233-9885-9a471d3d00e6.webm)

Only clicking I did here was selecting Portal 2, the workshop, and minimizing the workshop.

notfrying1pans:
I think it's an issue with how the ui shows up

slambo311:
> I think it's an issue with how the ui shows up

Yeah, that's pretty obvious.

Anyway, despite my lack of coding knowledge, I looked in the code myself (seeing as Jasper likely is sick of my crap), and yeah, couldn't find a thing.

PFedak:
Should be fixed now

PR #579, opened 2023-03-11
Cancel dragging over ScrollSelects if mouse buttons are up.

PFedak:
Various things can cause dragging to end without us registering a mouseup event, such as another window popping up, an exception or breakpoint, or leaving the window area. While this might be the right use case for ondragleave etc. I couldn't get that to work.

This just does the obvious thing and stops dragging if the mouse buttons are up, which handles the annoying cases like the Portal 2 workshop window without impacting normal dragging too much.

### DIFF
--- a/src/ui.ts
+++ b/src/ui.ts
@@ -358,6 +358,8 @@ export abstract class ScrollSelect implements Widget {
                     this.isDragging = false;
                 };
                 outer.onmouseover = (e) => {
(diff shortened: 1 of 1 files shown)


PFedak:
Seems to fix #575 and other random annoying things I hit when loading errors during development.

magcius:
Oh, nice, I don't think I knew about `e.buttons`. Thanks!

PR #582, opened 2023-03-20
Enhancement: Centralize Loading/Usage of Rust WASM

dcvz:
Centralises the loading and referencing of the Rust WASM lib. Atm its being loaded in specific places and requires that wherever you do load it is set up to be async. This moves loading to the main method (which is already async) and can be referenced now from all code. 

It simplifies some workarounds that were currently in the code. I'm not that familiar with web tech so I'm not sure if there's a performance hit of loading in main vs only when necessary.. but it seems like it was used in some hot paths likely to be hit anyway.


magcius:
I think back when Rust was a minor consideration, it made sense to load it at game-time, but now, this seems like a good clean-up. Thanks!

PR #601, opened 2023-08-01
getAnimFrame could get stuck in an infinite loop

Faraphel:
For some reason, when dropping custom files (such as Mario Kart Wii custom tracks) on the website, the page could freeze. It appear that the value "lastFrame" in "getAnimFrame" would be equal to 0, and if the loop mode is on repeat, the website would end up on an infinite loop.

Tested with this custom track : 
https://drive.google.com/file/d/1kttPIdbj6LF-Qdz9tFAbVC3k2YR9Lesv/view?usp=sharing


magcius:
Sure. I'd have to look into how the runtime handles this anyway, this animation logic is empirical anyway.

PR #608, opened 2023-08-16
Twilight Princess: Implement layer scenario switching

TakaRikka:
This changes the layer tab to a single select to have each layer be it's own scenario. For example this lets you switch to layer 14 on maps that have twilight to view the twilight version of the map, reinitializing the lighting appropriately. This also fixes bg_obj actors to be tied to the correct layer as their parent set_bg_obj actor


src/TwilightPrincess/ztp_scenes.ts
```diff
@@ -338,25 +343,52 @@ export class TwilightPrincessRenderer implements Viewer.SceneGfx {
...
+        console.log(`Current Layer: ${this.currentLayer}\nTwilight?: ${this.globals.world_dark}`);
+    }
+
     public createPanels(): UI.Panel[] {
         const getScenarioMask = () => {
             let mask: number = 0;
-            for (let i = 0; i < scenarioSelect.getNumItems(); i++)
-                if (scenarioSelect.itemIsOn[i])
-                    mask |= (1 << i);
+            mask |= (1 << this.scenarioSelect!.highlightedIndex);
```
magcius, line 372: this can be reworked, I'll handle it

PR #612, opened 2023-08-22
Twilight Princess: a bunch of various actor additions

TakaRikka:
This adds a bunch of various actors to improve accuracy of various areas, as well as converting a couple legacy actors into more formal d_a actors. The ztp extractor was changed slightly to overwrite the actor profile's process ID in the case of duplicates so that actors such as `firewood2` can spawn properly, so the extractor will need to be ran again.


src/TwilightPrincess/tools/ztp_extractor.ts
```diff
@@ -335,6 +335,7 @@ function extractProfiles(binaries: Binary[]) {
         let pcName = view.getUint16(0x08);
         if (pcNameOverride !== -1) {
             pcName = pcNameOverride;
+            view.setUint16(0x08, pcNameOverride);
```
magcius, line 338: Can I ask how you encountered this, and how the original game copes? I missed this logic when it first when in.

If we change this file (`f_pc_profiles.crg1_arc`, right?) then we're going to have to bump the cacheBust -- look at where it's requested in createScene and bump the second parameter there. I notice that the extra already has a cacheBust of 8 which is probably copied from zww. Feel free to remove that one again.

src/WindWaker/SComponent.ts
```diff
@@ -93,6 +125,10 @@ export function cM__Rad2Short(v: number): number {
     return v * (0x8000 / Math.PI);
 }
 
+export function cM__Deg2Short(v: number): number {
+    return v * 182.04445;
```
magcius, line 129: `return v * MathConstants.DEG_TO_RAD * kUshortTo2PI`. Not ideal we have two functions here for this between SComponent and m_do_mtx, oof. I can try to clean that up.

src/TwilightPrincess/d_kankyo.ts
```diff
@@ -559,6 +559,16 @@ function setLight_palno_get(dst: setLight_palno_ret, pselenvr: setLight_palno_ps
         currPselIdx = 0;
     }
 
+    if (prevPselIdx > envLight.colo.length) {
+        console.log(`setLight_palno_get: prevPselIdx (${prevPselIdx}) out of bounds! Colo entry num: ${envLight.colo.length}`);
```
magcius, line 563: Yeah, I noticed this kind of thing happened when I was messing around with toggling rooms in Hyrule Castle Rooftops.

src/TwilightPrincess/d_a.ts
```diff
@@ -1054,15 +1056,17 @@ class d_a_obj_glowSphere extends fopAc_ac_c {
...
         const l_colorCG = [0x96, 0x64, 0xFF, 0x96];
         const l_colorCB = [0x96, 0x64, 0x00, 0xFF];
         const l_colorC = (l_colorCR[this.type] << 24) | (l_colorCG[this.type] << 16) | (l_colorCB[this.type] << 8) | 0xFF;
 
         const color_c = colorNewFromRGBA8(l_colorC);
-        this.model.materialInstances[0].setColorOverride(ColorKind.C0, color_c);
+
+        const m0 = this.model.modelMaterialData.materialData![0].material;
+        colorCopy(m0.colorConstants[1], color_k);
+        colorCopy(m0.colorRegisters[1], color_c);
```
magcius, line 1069: Any reason for this instead of setColorOverride? setColorOverride should be used most of the time.

src/TwilightPrincess/d_a.ts
```diff
@@ -2510,6 +2514,488 @@ class d_a_obj_life extends daItemBase {
...
+
+        if (unk_6fa === 0) {
+            if ((this.parameters & 0xFF) === 0xFF) {
+                this.active = true;
+                this.setLight(globals.g_env_light);
+            }
+        } else if ((this.parameters & 0xFF) === 0xFF) {
+            this.active = true;
+            this.setLight(globals.g_env_light);
+        }
```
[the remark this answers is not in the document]
TakaRikka: yea i messed up the logic here, the else should be `!==`. the `(this.parameters & 0xFF)` is actually supposed to be a `dComIfGs_isSwitch` check where the parameter is the switchNo, but for simplicities sake i made it just check whether there was no switch set (0xFF) since we obv dont have switches setup here. maybe that's not the best solution here idk

src/TwilightPrincess/tools/ztp_extractor.ts
TakaRikka: tbh i'm not entirely sure how the original handles this cause it makes no sense with my current understanding of how the framework works. my first encounter with it was with `d_a_obj_firewood2`, where it shares the same process ID of 0x39 with `d_a_obj_Lv1Cdl00`. they're completely separate rels though, and don't seem to do any checks to load different types as far as im aware. there's only a handful of these occurrences though, with firewood2 being the most obvious one since most of the others are just tags

magcius:
OK, I'm going to merge now, remake the profiles, and bump the cache bust.

PR #613, opened 2023-09-01
Dragon Quest VIII: Journey of the Cursed King (PS2)

ehcya:
This pull request adds support for the PS2 version of Dragon Quest 8. Aside from the TIM2 texture specs, all the reverse engineering work has been done from scratch, on the EU version of the game. 
It also features the day night cycle and a script interpreter to have accurate NPC routines (with skeletal, visibility and facial animations also added). Options to change the current game progress and hour to check the different layouts are also implemented.

### DIFF
--- a/src/DragonQuest8/bundle.ts
+++ b/src/DragonQuest8/bundle.ts
@@ -1,0 +1,27 @@
+import ArrayBufferSlice from "../ArrayBufferSlice.js";
+import { readString } from "../util.js";
+
+export interface BundleResourceInfo {
+    magic: number;
+    offset: number;
+    size: number;
+}
+
+export function parseBundle(buffer: ArrayBufferSlice, globalOffset: number = 0): Map<String, BundleResourceInfo> {
+    const view = buffer.createDataView();
+    let offs = 0;
+    let offsetToNext = 1;
+    const rNameToInfo: Map<String, BundleResourceInfo> = new Map<String, BundleResourceInfo>();
+
+    while (offsetToNext > 0) {
+        const rName: String = readString(buffer, offs);
+        const rSize: number = view.getInt32(offs + 0x44, true);
+        offsetToNext = view.getInt32(offs + 0x48, true);
+        if (offsetToNext > 0) {
+            const rMagic = view.getUint32(offs + 0x50, true);
+            rNameToInfo.set(rName, { magic: rMagic, offset: offs + 0x50 + globalOffset, size: rSize });
+        }
+        offs += offsetToNext;
+    }
+    return rNameToInfo;
+}
--- a/src/DragonQuest8/bundle.ts
+++ b/src/DragonQuest8/bundle.ts
@@ -11,7 +11,7 @@ export function parseBundle(buffer: ArrayBufferSlice, globalOffset: number = 0):
     const view = buffer.createDataView();
     let offs = 0;
     let offsetToNext = 1;
-    const rNameToInfo: Map<String, BundleResourceInfo> = new Map<String, BundleResourceInfo>();
+    const rNameToInfo = new Map<String, BundleResourceInfo>();
 
     while (offsetToNext > 0) {
         const rName: String = readString(buffer, offs);
--- a/src/DragonQuest8/render.ts
+++ b/src/DragonQuest8/render.ts
@@ -15,7 +15,7 @@ import { GfxBlendFactor, GfxBlendMode, GfxBuffer, GfxBufferUsage, GfxDevice, Gfx
 import { FormatCompFlags, FormatFlags, FormatTypeFlags, makeFormat } from '../gfx/platform/GfxPlatformFormat.js';
 import { GfxRenderCache } from '../gfx/render/GfxRenderCache.js';
 import { GfxRenderInst, GfxRenderInstManager, GfxRendererLayer, makeSortKey, setSortKeyDepth, setSortKeyLayer } from '../gfx/render/GfxRenderInstManager.js';
-import { nArray } from '../util.js';
+import { fallbackUndefined, nArray } from '../util.js';
 import * as Viewer from '../viewer.js';
 import * as CHR from './chr.js';
 import * as IMG from './img.js';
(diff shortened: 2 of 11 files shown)


src/DragonQuest8/render.ts
```diff
@@ -0,0 +1,683 @@
...
+    public bIsSkybox: boolean = false;
+    public mdtSubmeshInstances: MDTSubmeshInstance[] = [];
+    public debugJoints: boolean = false;
+    public IsRigidJointIDVisible: Map<number, boolean> = new Map<number, boolean>;
+    public jointMatrices: mat4[] = [];
+    public motion: MOT.Motion | null = null;
+    public texAnims: (IMG.TexAnim | null)[] = [];
+    public script: STB.STB | null = null;
+    public lastTick: number = 0;
+    public tickRateMs: number = 1 / 30 * 1000
```
magcius, line 351: here's a place you forgot a semicolon i guess

src/DragonQuest8/render.ts
```diff
@@ -0,0 +1,683 @@
...
+                // drawWorldSpacePoint(ctx, viewerInput.camera.clipFromWorldMatrix, scratchVec3b, Yellow,100);
+            }
+        }
+
+        const template = renderInstManager.pushTemplateRenderInst();
+        for (let i = 0; i < this.mdtSubmeshInstances.length; i++) {
+            if (this.mdtSubmeshInstances[i].bNeverVisible)
+                this.mdtSubmeshInstances[i].bIsVisible = false;
+            else
+                this.mdtSubmeshInstances[i].bIsVisible = this.IsRigidJointIDVisible.get(this.mdtSubmeshInstances[i].rigidJointId) as boolean;
```
magcius, line 481: a bit weird that you always set bIsVisible and all it does is skip something you can always do. How about:

```cpp
const submesh = this.mdtSubmeshInstances[i];
if (submesh.bNeverVisible)
    continue;
const jointVisible = this.IsRigidJointIDVisible.get(submesh.rigidJointID)!;
if (!jointVisible)
    continue;

...
```

src/DragonQuest8/bundle.ts
```diff
@@ -0,0 +1,27 @@
...
+    magic: number;
+    offset: number;
+    size: number;
+}
+
+export function parseBundle(buffer: ArrayBufferSlice, globalOffset: number = 0): Map<String, BundleResourceInfo> {
+    const view = buffer.createDataView();
+    let offs = 0;
+    let offsetToNext = 1;
+    const rNameToInfo: Map<String, BundleResourceInfo> = new Map<String, BundleResourceInfo>();
```
magcius, line 14: You can leave the type declaration off if it's "obvious" (e.g. already has a new expression on the rhs)

src/DragonQuest8/render.ts
```diff
@@ -0,0 +1,683 @@
...
+    }
+
+    public setVisible(visible: boolean): void {
+        this.bIsVisible = visible;
+    }
+
+    public destroy(device: GfxDevice): void {
+        for (let i = 0; i < this.mdtSubmeshInstances.length; i++)
+            this.mdtSubmeshInstances[i].destroy(device);
+        this.mdsData.destroy(device);
```
magcius, line 502: This usual reason I have the Data/Instance split is so that I can have multiple copies of the same MDSData (e.g. twelve copies of the same crate), so having the Instance destroy the Data seems a bit counterproductive, since that implies it's always owning a unique copy of the data. This is usually where stuff like the ModelCache comes in -- it stores the Data and then Instances are created off of that.

magcius (commented):
Lots of comments here. All of these are suggestions -- I'll merge even if none of them are fixed, but I did a cursory review pass. You decide if you want to fix things!

General suggestions:
* You do a lot of `objArr[i]` indexing inline. I find the code is nicer when you pull that indexed value out into its own variable -- it helps separate the concepts of "this is a thing I need" and "the thing I need is inside this array".
* You use a lot of Maps, too. And I think that's fine, but again, I think the code is nicer if you can isolate the Map parts to their own little section of the code. Helper functions are great, use more of them :)

src/DragonQuest8/render.ts
ehcya: Yes, I was actually planning to use this logic for duplicate instances (NPCs or props) but the animation data made it hard to track everything so I first went with this 1:1 equivalent. I'll revisit this later and will take a look at the ModelCache.

src/DragonQuest8/render.ts
```diff
@@ -0,0 +1,683 @@
...
+    public tickRateMs: number = 1 / 30 * 1000;
+
+
+    constructor(cache: GfxRenderCache, public chrs: CHR.CHR[], public transforms: mat4[], public eulerRotations: vec3[], public chrNPCDayPeriods: (SINFO.ENPCDayPeriod | null)[], public chrDayPeriodFlags: (number | null)[], public stbs: (STB.STB | null)[] | null = null, public chrProgressFlags: (number | null)[]) {
+        this.name = "CHR renderers";
+        for (let i = 0; i < chrs.length; i++) {
+            const chr = chrs[i];
+            this.motNameToMotionMaps.push(chr.mot !== null ? chr.mot.motionNameToMotion : null);
+            if (chr.model !== null) {
+                this.MDSRenderers.push(new MDSInstance(cache, new MDSData(cache, chr.model), transforms[i], eulerRotations[i], chr.img, chr.mot, chr.model.name, chrNPCDayPeriods[i], chrDayPeriodFlags[i], chrProgressFlags[i]));
```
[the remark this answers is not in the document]
ehcya: By construction the chrs and the other lists will always have the same length (with dummy entries if necessary) so that should work out just fine,

ehcya:
Thank you for the great suggestions, as this is my first time writing typescript they're very welcome. I implemented all of these, please let me know if there's anything left you want me to change.

src/DragonQuest8/scenes.ts
```diff
@@ -0,0 +1,688 @@
...
+            "p024d.chr", 
+            "p024c.chr", 
+            "p024d.chr", 
+            "p024e.chr", 
+            "p073a.chr", 
+            "p073a_iro.chr",
+            "sp001.chr", 
+        ]);
+
+        const dispoBuffer = await dataFetcher.fetchData("DragonQuest8/event/villager/disposition/" + this.mID + ".cfg", { allow404: true });
```
magcius, line 218:
```ts
await dataFetcher.fetchData(`DragonQuest8/event/villager/disposition/${this.mID}.cfg`, { allow404: true })
```

and other places you use `"foo" + bar` on strings, I recommend using format strings instead.

Also, should this (and the below) have the allow404 flag?

src/DragonQuest8/stb.ts
```diff
@@ -0,0 +1,1024 @@
...
+import { computeModelMatrixSRT, getMatrixTranslation, setMatrixTranslation } from '../MathHelpers.js';
+import { assert, readString } from "../util.js";
+import * as IMG from './img.js';
+import * as MOT from './mot.js';
+import { MDSInstance } from './render.js';
+
+//Misc, hardcoded for now
+const MAIN_PROGRESS_ID = 0x1;
+const SUB_PROGRESS_ID = 0x0;
+const EVENT_FLAGS = 0x0;
```
magcius, line 12: Should these be in the StageInfo struct?

src/DragonQuest8/render.ts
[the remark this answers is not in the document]
magcius: Then remove the `chr.model !== null` check

magcius (approved):

src/DragonQuest8/scenes.ts
ehcya: Yes the flag should be there as some areas don't have NPCs.

src/DragonQuest8/stb.ts
ehcya: Indeed, it would make more sense.

PR #629, opened 2024-02-25
Battle for Bikini Bottom: Rewrite renderer from scratch

seilweiss:
This replaces the Battle for Bikini Bottom renderer with an improved one that is more accurate to the original game. Many of the more noticeable issues with the current renderer (transparency sorting, missing textures, models disappearing that shouldn't, etc.) have been fixed.

The new renderer has about the same features as the current one, so no animations, moving platforms, etc. yet. I'll be working on adding those soon.

Patrick's unused dream level is removed for now since it uses an older level format and DXT compressed textures which I don't support yet.


src/HeavyIron/rw/pipelines/AtomicAllInOne.ts
```diff
@@ -0,0 +1,351 @@
...
+layout(std140) uniform ub_AtomicParams {
+    Mat4x4 u_Projection;
+    Mat4x4 u_ModelView;
+    DirectionalLight u_DirectionalLights[MAX_DIRECTIONAL_LIGHTS];
+    vec4 u_AmbientColor;
+    vec4 u_FogColor;
+    float u_FarPlane;
+    float u_FogPlane;
+    float u_AlphaRef;
+    float u_EnablePrelit;
```
magcius, line 53: Be careful about using lone floats, especially around other vectors, as the rules for their packing can be fairly unobvious. In other renderers, I usually do a pattern like:

```glsl
layout(std140) uniform Thing {
    vec4 u_Misc[1];
};

#define u_FarPlane (u_Misc[0].x)
#define u_FogPlane (u_Misc[0].y)
```

What you have here *should* work, but just be careful.

src/HeavyIron/rw/pipelines/AtomicAllInOne.ts
```diff
@@ -0,0 +1,351 @@
...
+
+        const directionalColor = scratchColor;
+        const ambientColor = scratchColor2;
+
+        colorCopy(directionalColor, OpaqueBlack);
+        colorFromRGBA(ambientColor, 0, 0, 0, (geom.flags & RpGeometryFlag.LIGHT) ? 1.0 : 0.0);
+
+        let lightCount = 0;
+        if (atomic.geometry.flags & RpGeometryFlag.LIGHT) {
+            for (const light of rw.world.lights) {
```
magcius, line 259: This isn't too much math, but since this is global to the world, I wonder if it would make sense to just do this once, and shove it in the scene globals, unless you were thinking of adding local lights later.

Also, a modern engine does lighting in world-space, in which case a lot of this math simply goes away.

src/HeavyIron/rw/rwcore.ts
```diff
@@ -0,0 +1,778 @@
...
+    VMASK = 0xF000,
+    MASK = (UMASK | VMASK)
+}
+
+export const enum RwTextureStreamFlags {
+    NATEXTURESTREAMFLAG = 0x00,
+    USERMIPMAPS = 0x01
+}
+
+// Not sure if these are right
```
magcius, line 312: * NEAREST = GfxMipFilterMode.NoMip, GfxTexFilterMode.Point
* LINEAR = GfxMipFilterMode.NoMip, GfxTexFilterMode.Bilinear
* MIPNEAREST = GfxMipFilterMode.Nearest, GfxTexFilterMode.Point
* MIPLINEAR = GfxMipFilterMode.Nearest, GfxTexFilterMode.Bilinear
* LINEARMIPNEAREST = GfxMipFilterMode.Linear, GfxTexFilterMode.Point
* LINEARMIPLINEAR = GfxMipFilterMode.Linear, GfxTexFilterMode.Bilinear

src/HeavyIron/rw/rwcore.ts
```diff
@@ -0,0 +1,778 @@
...
+        return this.gfxSampler;
+    }
+
+    public get filter() {
+        return this._filter;
+    }
+
+    public set filter(f: RwTextureFilterMode) {
+        this._filter = f;
+        this.gfxSampler = null;
```
magcius, line 382: Is it realistic to change these at runtime?

src/HeavyIron/rw/rwcore.ts
```diff
@@ -0,0 +1,778 @@
...
+    case RwBlendFunction.INVSRCALPHA:  return GfxBlendFactor.OneMinusSrcAlpha;
+    case RwBlendFunction.DESTALPHA:    return GfxBlendFactor.DstAlpha;
+    case RwBlendFunction.INVDESTALPHA: return GfxBlendFactor.OneMinusDstAlpha;
+    case RwBlendFunction.DESTCOLOR:    return GfxBlendFactor.Dst;
+    case RwBlendFunction.INVDESTCOLOR: return GfxBlendFactor.OneMinusDst;
+    case RwBlendFunction.SRCALPHASAT:  return GfxBlendFactor.SrcAlpha; // unsupported
+    }
+}
+
+function convertGfxBlendFactor(factor: GfxBlendFactor): RwBlendFunction {
```
magcius, line 577: I would probably just store the Rw variants of enums and convert to Gfx in the renderer.

src/HeavyIron/rw/rwcore.ts
```diff
@@ -0,0 +1,778 @@
...
+    switch (cull) {
+    case GfxCullMode.None:         return RwCullMode.NONE;
+    case GfxCullMode.Back:         return RwCullMode.BACK;
+    case GfxCullMode.Front:        return RwCullMode.FRONT;
+    case GfxCullMode.FrontAndBack: return RwCullMode.NONE; // unsupported
+    }
+}
+
+export class RwRenderState {
+    public megaStateFlags: Partial<GfxMegaStateDescriptor> = makeMegaState();
```
magcius, line 610: This method is OK, but where possible, I prefer caching things up-front, in case you need to generate shader variants depending on the render settings. Though in your case, branching on uniforms is probably good enough for all the cases you might need to handle (vertexAlphaEnable)

src/HeavyIron/rw/pipelines/AtomicAllInOne.ts
```diff
@@ -0,0 +1,351 @@
...
+
+void main() {
+    gl_Position = Mul(u_Projection, Mul(u_ModelView, vec4(a_Position, 1.0)));
+
+    // Normal (in eye space)
+    vec3 t_Normal = normalize(Mul(u_ModelView, vec4(a_Normal, 0.0)).xyz);
+
+    vec4 t_Color = u_EnablePrelit != 0.0 ? a_Color : (u_EnableLight != 0.0 ? vec4(0, 0, 0, 1) : vec4(1.0));
+
+    if (u_EnableLight != 0.0) {
```
magcius, line 90: If you need to pack a lot of flags into a single float, you can cast the float field to an integer, and then test bitfield positions. See e.g. https://github.com/magcius/noclip.website/blob/master/src/DiddyKongRacing/F3DDKR_Program.ts#L36-L49

src/HeavyIron/rw/pipelines/AtomicAllInOne.ts
```diff
@@ -0,0 +1,351 @@
...
+        if (atomic.geometry.flags & RpGeometryFlag.LIGHT) {
+            for (const light of rw.world.lights) {
+                if (light.type === RpLightType.DIRECTIONAL) {
+                    if (lightCount < MAX_DIRECTIONAL_LIGHTS) {
+                        const mat = scratchMat4;
+                        mat4.mul(mat, rw.camera.viewMatrix, light.frame.matrix);
+
+                        const dir = scratchVec4;
+                        vec4.set(dir, mat[8], mat[9], mat[10], 0);
+                        vec4.normalize(dir, dir);
```
magcius, line 267: Using a vec3 and getMatrixAxisZ() might be a bit easier here.

PR #631, opened 2024-02-26
Battle for Bikini Bottom: RenderWare improvements

seilweiss:
This addresses/implements the suggestions from #629.

- Lone float uniforms are now grouped together in a vec4 to avoid float packing issues.
- Lighting state is now cached as part of the atomic pipeline shared across all RpAtomic instances. All lights are static in BFBB, so currently the cache is only rebuilt whenever any lights have been enabled/disabled or added/removed from the world (this does happen a handful of times per frame).
- Lighting is done in world-space instead of eye-space now. This eliminates having to transform the lights into eye-space, but it does add an extra matrix multiplication to the shader (model and view matrices have to be separate now). I'm not quite sure if this is faster.
- Texture filter modes now convert to the correct Gfx equivalents.
- It's possible to change the filter/address mode on textures at runtime, but afaik BFBB doesn't do this, so I've removed the code that supports this for now.
- RwRenderState only stores Rw variants of enums now and the conversion to Gfx is done in rendering pipelines. RwRenderState still maintains its own texture filter/address mode as well as a reference to a RwRaster, to be used by the immediate mode pipeline (plan to implement soon).
- I may consider generating shader variants instead of branching on uniforms soon, especially when it comes time to support changing the alpha test comparison operator (I don't imagine branching on an enum in the fragment shader would be very efficient).
- Bitflags are now packed into a single float in the uniform buffers.
- I'm now using a vec3 and getMatrixAxisZ() for the light direction caching.

I want to separate more Gfx stuff from RenderWare's public types, e.g. RwTexture/RwRaster should not contain a GfxTexture/GfxSampler/etc. I'm considering adding a "RwGfx" helper class to RwEngine that holds all the Gfx-specific data and helper functions that can be reused across multiple RenderWare pipelines.

### DIFF
--- a/src/HeavyIron/HIEntPickup.ts
+++ b/src/HeavyIron/HIEntPickup.ts
@@ -83,7 +83,7 @@ export class HIEntPickupManager {
     }
     
     public render(scene: HIScene, rw: RwEngine) {
-        scene.lightKitManager.enable(null, rw.world, scene);
+        scene.lightKitManager.enable(null, scene);
 
         const src = this.pickupOrientation;
         for (const pkup of this.pickups) {
--- a/src/HeavyIron/HILightKit.ts
+++ b/src/HeavyIron/HILightKit.ts
@@ -1,5 +1,5 @@
 import { mat4 } from "gl-matrix";
-import { RpLight, RpLightType, RpWorld } from "./rw/rpworld.js";
+import { RpLight, RpLightFlag, RpLightType } from "./rw/rpworld.js";
 import ArrayBufferSlice from "../ArrayBufferSlice.js";
 import { RwEngine, RwStream } from "./rw/rwcore.js";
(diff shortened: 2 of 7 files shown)


seilweiss:
Added a new commit, per our Discord discussion I removed the light tracking fom the global atomic pipeline for now and lights are now recalculated per RpAtomic instance again. I wonder if it would be worth experimenting caching the lights on each RpAtomic instance since they usually don't change between renders of the same instance.

I also moved megaStateFlags to be per RpAtomic instance instead of in the global pipeline. Similarly to lights, I'm considering keeping a hash of RwRenderState in each instance as it usually doesn't change between renders, so I might be able to avoid redoing the conversion to megaStateFlags every render. This might be a negligible optimization though.

Happy to continue adding to this PR if you have any concerns/suggestions.

PR #672, opened 2024-05-07
misc wow fixes

wgreenberg:
* handle deep/shallow rivers and oceans, use correct liquid transparency values
* make particle volumes invisible

### DIFF
--- a/rust/src/wow/wmo.rs
+++ b/rust/src/wow/wmo.rs
@@ -452,7 +452,7 @@ impl WmoLiquid {
                 vertex_prototypes.push([
                     pos_x, pos_y, pos_z,
                     x as f32, y as f32,
-                    1.0, 0.0, // deep, fishable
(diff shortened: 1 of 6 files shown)


rust/src/wow/wmo.rs
```diff
@@ -452,7 +452,7 @@ impl WmoLiquid {
                 vertex_prototypes.push([
                     pos_x, pos_y, pos_z,
                     x as f32, y as f32,
-                    1.0, 1.0, // deep, fishable
+                    0.0, 1.0, // deep, fishable
```
magcius, line 455: Is this intended to revert deep water?

wgreenberg: yeah, afaict "deep" water is only used for oceans, and any ocean tile in a WMO shouldn't be deep afaik

PR #678, opened 2024-06-05
wow: ADT shadowmaps

wgreenberg:


src/WorldOfWarcraft/program.ts
```diff
@@ -757,6 +759,7 @@ void mainPS() {
       vec3(0.0), // specular
       vec3(0.0) // emissive
     ), 1.0);
+    finalColor.rgb = finalColor.rgb * (1.0 - shadow * 0.333);
```
magcius, line 762: Hmm, is this coming from somewhere, or is this a guess? From what I can tell from the GLSL shaders, the shadow is multiplied into lDiffuse, like `lDiffuse = (exteriorDirectColor.xyz * nDotL) * shadow;`

wgreenberg: it's based on noggit, which is a WoW map editor, but i suspect their implementation is a total guess. scaling down the diffuse looks pretty good though -- there's also an unused but suspicious part of the global light data called `shadowOpacity`, which is actually a full color. do you see anything that might resemble that in the GLSL shaders?

wgreenberg: i'll double check the client to see if it's like this

magcius: It seems like they multiply the shadow contribution into lSpecular as well. This is based on the client shaders sampling pc_shadowMaskTex. That has shadows packed in R, and SSAO packed in G, so I imagine it's a runtime shadow map thing; no clue if it has the static shadow maps baked in it as well.

wgreenberg: nice, yeah that looks way better. pushed a commit that scales specular too

src/WorldOfWarcraft/program.ts
```diff
@@ -27,7 +27,8 @@ vec3 calcLight(
   vec3 accumLight,
   vec3 precomputedLight,
   vec3 specular,
-  vec3 emissive) {
+  vec3 emissive,
+  float shadow) {
```
magcius, line 31: At some point we should probably move calcLight to take a struct as input, or clean it up in some other ways.

PR #680, opened 2024-06-07
wow: Fix WMOs with shared doodads

wgreenberg:
The waterwheel in Warsong Gulch is currently all-black due to the weird case of WMO groups which share the same doodad. I couldn't find any docs on how this is handled, and didn't really see where WowWebViewer handles it, so simply choose the group that's closest to the doodad as its group for the purposes of lighting.

### DIFF
--- a/src/WorldOfWarcraft/program.ts
+++ b/src/WorldOfWarcraft/program.ts
@@ -257,7 +257,8 @@ layout(std140) uniform ub_ModelParams {
 
 layout(std140) uniform ub_BatchParams {
     vec4 shaderParams; // vertexShader, pixelShader, _, _
-    vec4 materialParams; // blendMode, applyInteriorLight, applyExteriorLight, _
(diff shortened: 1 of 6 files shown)


src/WorldOfWarcraft/program.ts
```diff
@@ -257,7 +257,8 @@ layout(std140) uniform ub_ModelParams {
 
 layout(std140) uniform ub_BatchParams {
     vec4 shaderParams; // vertexShader, pixelShader, _, _
-    vec4 materialParams; // blendMode, applyInteriorLight, applyExteriorLight, _
+    vec4 materialParams; // blendMode, applyInteriorLight, applyExteriorLight, unlit
+    vec4 moreMaterialParams; // unfogged, exterior_light, sidn, window
```
magcius, line 261: Hmm, I wonder if it would make some sense to switch to bitflags for this. Not a big deal though.

PR #688, opened 2024-07-04
Add wow particles

wgreenberg:
sorry jasper

### DIFF
--- a/rust/src/wow/common.rs
+++ b/rust/src/wow/common.rs
@@ -92,7 +92,7 @@ impl<'a> Iterator for ChunkedData<'a> {
 pub type WowCharArray = WowArray<u8>;
 
 impl WowArray<u8> {
-    pub fn to_string(&self, data: &[u8]) -> Result<String, DekuError> {
+    pub fn to_string(&self, data: &[u8]) -> Result<String, String> {
         let mut bytes = self.to_vec(data)?;
         bytes.pop(); // pop the null byte
         Ok(String::from_utf8(bytes).unwrap())
@@ -243,15 +243,16 @@ pub struct WowArray<T> {
     pub count: u32,
     pub offset: u32,
     #[deku(skip)]
-    element_type: std::marker::PhantomData<T>,
+    pub element_type: std::marker::PhantomData<T>,
 }
 
 impl<T> WowArray<T> where for<'a> T: DekuRead<'a> {
-    pub fn to_vec(&self, data: &[u8]) -> Result<Vec<T>, DekuError> {
+    pub fn to_vec(&self, data: &[u8]) -> Result<Vec<T>, String> {
         let mut result = Vec::with_capacity(self.count as usize);
         let mut bitslice = BitSlice::from_slice(&data[self.offset as usize..]);
         for _ in 0..self.count {
-            let (new_bitslice, element) = T::read(bitslice, ())?;
+            let (new_bitslice, element) = T::read(bitslice, ())
+                .map_err(|e| format!("{:?}", e))?;
             bitslice = new_bitslice;
             result.push(element);
         }
--- a/rust/src/wow/common.rs
+++ b/rust/src/wow/common.rs
@@ -2,6 +2,23 @@ use deku::{bitvec::{BitSlice, BitVec}, ctx::ByteSize, prelude::*};
 use wasm_bindgen::prelude::*;
 use std::ops::{Mul, AddAssign};
 
+#[derive(DekuRead, Debug, Clone, Copy)]
+pub struct Fixedi16 {
+    pub inner: i16,
+}
+
+impl From<Fixedi16> for f32 {
+    fn from(value: Fixedi16) -> Self {
+        value.inner as f32 / 32768.0
+    }
+}
+
+impl From<f32> for Fixedi16 {
+    fn from(value: f32) -> Self {
+        Self { inner: (value * 32768.0) as i16 }
+    }
+}
+
 #[derive(DekuRead, Debug)]
 pub struct Chunk {
     pub magic: [u8; 4],
@@ -97,6 +114,14 @@ impl WowArray<u8> {
         bytes.pop(); // pop the null byte
         Ok(String::from_utf8(bytes).unwrap())
     }
+}
+
+pub fn fixed_precision_6_9_to_f32(x: u16) -> f32 {
+    let mut result = (x & 0x1ff) as f32 * (1.0 / 512.0) + (x >> 9) as f32;
+    if x & 0x8000 > 0 {
+        result *= -1.0;
+    }
+    result
 }
 
 #[wasm_bindgen(js_name = "WowQuat")]
@@ -240,8 +265,8 @@ impl AABBox {
 
 #[derive(Debug, DekuRead, Clone, Copy)]
 pub struct WowArray<T> {
-    pub count: u32,
-    pub offset: u32,
+    pub count: i32,
+    pub offset: i32,
     #[deku(skip)]
     pub element_type: std::marker::PhantomData<T>,
 }
@@ -307,9 +332,32 @@ impl Lerp for u8 {
     }
 }
 
+impl Lerp for Fixedi16 {
+    fn lerp(self, other: Self, t: f32) -> Self {
+        let a: f32 = self.into();
+        let b: f32 = other.into();
+        Fixedi16::from(a.lerp(b, t))
+    }
+}
+
+impl Lerp for i16 {
+    fn lerp(self, other: Self, t: f32) -> Self {
+        ((self as f32) * (1.0 - t) + (other as f32) * t) as i16
+    }
+}
+
 impl Lerp for u16 {
     fn lerp(self, other: Self, t: f32) -> Self {
         ((self as f32) * (1.0 - t) + (other as f32) * t) as u16
+    }
+}
+
+impl Lerp for Vec2 {
+    fn lerp(self, other: Self, t: f32) -> Self {
+        Vec2 {
+            x: self.x * (1.0 - t) + other.x * t,
+            y: self.y * (1.0 - t) + other.y * t,
+        }
     }
 }
 
--- a/src/WorldOfWarcraft/data.ts
+++ b/src/WorldOfWarcraft/data.ts
@@ -426,16 +426,20 @@ export class ModelData {
   }
 
   public updateAnimation(view: View) {
-    this.animationManager.update_animations(
-      view.deltaTime,
+    this.animationManager.update(view.deltaTime);
+    this.animationManager.update_textures(
       this.textureWeights,
       this.textureTranslations,
       this.textureRotations,
       this.textureScalings,
+    );
+    this.animationManager.update_bones(
       this.boneTranslations,
       this.boneRotations,
       this.boneScalings,
-      this.vertexColors,
+    );
+    this.animationManager.update_vertex_colors(this.vertexColors);
+    this.animationManager.update_lights(
       this.ambientLightColors,
       this.diffuseLightColors,
       this.lightAttenuationStarts,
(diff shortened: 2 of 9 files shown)


rust/src/wow/common.rs
```diff
@@ -92,13 +109,21 @@ impl<'a> Iterator for ChunkedData<'a> {
...
 impl WowArray<u8> {
-    pub fn to_string(&self, data: &[u8]) -> Result<String, DekuError> {
+    pub fn to_string(&self, data: &[u8]) -> Result<String, String> {
         let mut bytes = self.to_vec(data)?;
         bytes.pop(); // pop the null byte
         Ok(String::from_utf8(bytes).unwrap())
     }
 }
 
+pub fn fixed_precision_6_9_to_f32(x: u16) -> f32 {
```
magcius, line 119: nice

src/WorldOfWarcraft/data.ts
```diff
@@ -303,6 +304,532 @@ export class SkyboxData {
   }
 }
 
+function uniform(min: number, max: number): number {
+  return lerp(min, max, Math.random());
```
magcius, line 308: this is verbatim randomRange from MathHelpers.ts

src/WorldOfWarcraft/data.ts
```diff
@@ -303,6 +304,532 @@ export class SkyboxData {
...
+  public alive = true;
+  public color = vec4.create();
+  public scale = vec2.create();
+  public texCoordHead = vec2.create();
+  public texCoordTail = vec2.create();
+
+  static createSpline(emitter: ParticleEmitter, dt: number): Particle {
+    let areaX = clamp(emitter.emissionAreaWidth, 0, 1);
+    let t = clamp(emitter.emissionAreaLength, 0, 1);
+    console.warn(`SPLINE`)
```
magcius, line 322: same tbh

src/WorldOfWarcraft/data.ts
```diff
@@ -303,6 +304,532 @@ export class SkyboxData {
...
+  public age = 0;
+  public alive = true;
+  public color = vec4.create();
+  public scale = vec2.create();
+  public texCoordHead = vec2.create();
+  public texCoordTail = vec2.create();
+
+  static createSpline(emitter: ParticleEmitter, dt: number): Particle {
+    let areaX = clamp(emitter.emissionAreaWidth, 0, 1);
+    let t = clamp(emitter.emissionAreaLength, 0, 1);
```
magcius, line 321: clamp(x, 0, 1) is also the same as saturate(x) from MathHelpers.ts

src/WorldOfWarcraft/data.ts
```diff
@@ -303,6 +304,532 @@ export class SkyboxData {
...
+      velocity[2] = dz;
+      vec3.normalize(velocity, velocity);
+      vec3.scale(velocity, velocity, emitter.getEmissionSpeed());
+    } else if (emitter.verticalRange !== 0.0) {
+      // this is insane. treat the spline's derivative at t as a rotation vector, and the
+      // emitter's verticalRange parameter as a rotation (in degrees). then, set the velocity
+      // to the resulting rotation along just the Z axis. i guess.
+      const rotAxis = emitter.spline.calculateParametricSplineDerivative(vec3.create(), t);
+      vec3.normalize(rotAxis, rotAxis);
+      const rotRadians = emitter.verticalRange * Math.PI / 180.0;
```
magcius, line 350: MathConstants.DEG_TO_RAD

src/WorldOfWarcraft/data.ts
```diff
@@ -303,6 +304,532 @@ export class SkyboxData {
...
+      const rotQuat = quat.setAxisAngle(quat.create(), rotAxis, uniform(-1, 1) * rotRadians);
+      const rotMat = mat3.fromQuat(mat3.create(), rotQuat);
```
magcius, line 352: I think this could be `mat4.fromRotation` instead of needing the quat but whatever

src/WorldOfWarcraft/data.ts
```diff
@@ -303,6 +304,532 @@ export class SkyboxData {
...
+    if (emitter.zSource > 0.001) {
+      const dz = position[2] - emitter.zSource;
+      velocity = vec3.clone(position);
+      velocity[2] = dz;
+      vec3.normalize(velocity, velocity);
+      vec3.scale(velocity, velocity, emitter.getEmissionSpeed());
+    } else if (emitter.verticalRange !== 0.0) {
+      // this is insane. treat the spline's derivative at t as a rotation vector, and the
+      // emitter's verticalRange parameter as a rotation (in degrees). then, set the velocity
+      // to the resulting rotation along just the Z axis. i guess.
```
magcius, line 347: very strange, I tried to come up with a better geometric interpretation and I couldn't. All I can say is that the derivative is going to be the travel direction along the spline, aka the "facing" direction, but that doesn't explains the Z axis stuff. I suspect the name "verticalRange" is mostly from its use in another emitter, in the original code it might have just been a union, or was called "param1" or something

src/WorldOfWarcraft/data.ts
```diff
@@ -303,6 +304,532 @@ export class SkyboxData {
...
+      this.verticalRange,
+      this.horizontalRange,
+      this.lifespan,
+      this.emissionRate,
+      this.emissionAreaLength,
+      this.emissionAreaWidth,
+      this.zSource,
+    ] = this.updateBuffer;
+    if (this.emitter.use_compressed_gravity()) {
+      console.warn(`compressed!!`)
```
magcius, line 664: same

src/WorldOfWarcraft/data.ts
```diff
@@ -303,6 +304,532 @@ export class SkyboxData {
...
+      material0x01 = false;
+      material0x20 = this.emitter.check_flag(0x40000000);
+    } else if (this.emitter.check_flag(0x100000)) {
+      material0x01 = false;
+    } else {
+      material0x01 = !this.emitter.check_flag(0x1);
+    }
+
+    const multiTex = this.emitter.check_flag(0x10000000);
+    if (this.particleType === 2 || (this.particleType === 4 && multiTex) && this.txac !== 0) {
```
magcius, line 633: double-check these parens, not sure if this is meant to be `(particleType === 2 || (particleType === 4 && multiTex)) && this.txac !== 0`, or `particleType === 2 || ((particleType === 4 && multiTex) && this.txac !== 0)`.

i suspect the latter because of the `particleType === 2` below, but i'm not sure if it will parse that way.

src/WorldOfWarcraft/data.ts
```diff
@@ -303,6 +304,532 @@ export class SkyboxData {
...
+      particle = Particle.createSpline(this, dt);
+    } else {
+      throw new Error(`unknown particle emitter type ${this.emitter.emitter_type}`);
+    }
+
+    if (!this.emitter.check_flag(0x10)) {
+      vec3.transformMat4(particle.position, particle.position, this.modelMatrix);
+      const transformedVelocity = vec4.fromValues(particle.velocity[0], particle.velocity[1], particle.velocity[2], 0);
+      vec4.transformMat4(transformedVelocity, transformedVelocity, this.modelMatrix);
+      vec3.set(particle.velocity, transformedVelocity[0], transformedVelocity[1], transformedVelocity[2]);
```
magcius, line 716:
```suggestion
      transformVec3Mat4w0(particle.velocity, this.modelMatrix, particle.velocity);
```

src/WorldOfWarcraft/data.ts
```diff
@@ -303,6 +304,532 @@ export class SkyboxData {
...
+        particle.scale[0] = this.updateBuffer[4];
+        particle.scale[1] = this.updateBuffer[5];
+        const cellHead = this.updateBuffer[6];
+        this.extractTexCoords(particle.texCoordHead, cellHead);
+        const cellTail = this.updateBuffer[6];
+        this.extractTexCoords(particle.texCoordTail, cellTail);
+
+        vec3.scaleAndAdd(particle.velocity, particle.velocity, this.force, dtSeconds);
+        if (this.emitter.drag > 0) {
+          vec3.scale(particle.velocity, particle.velocity, 1.0 - this.emitter.drag * dtSeconds);
```
magcius, line 797: for scales you usually want `(1.0 - emitter.drag) ** dtSeconds` for the right integration usually

src/WorldOfWarcraft/data.ts
```diff
@@ -303,6 +304,532 @@ export class SkyboxData {
...
+    const scratchVec3 = vec3.create();
+    for (let i=0; i<this.particles.length; i++) {
+      const particle = this.particles[i];
+      let offs = ParticleEmitter.TEXELS_PER_PARTICLE * i * 4;
+      vec3.copy(scratchVec3, particle.position);
+      if (this.emitter.translate_particle_with_bone()) {
+        vec3.transformMat4(scratchVec3, scratchVec3, this.modelMatrix);
+      }
+      // offs += fillVec4(pixels, offs, particle.position[0], particle.position[1], particle.position[2]);
+      offs += fillVec4(pixels, offs, scratchVec3[0], scratchVec3[1], scratchVec3[2]);
```
magcius, line 823: fillVec3v

src/WorldOfWarcraft/data.ts
```diff
@@ -303,6 +304,532 @@ export class SkyboxData {
...
+      offs += fillVec4v(pixels, offs, particle.color);
+      offs += fillVec4(pixels, offs, particle.scale[0], particle.scale[1]);
+      offs += fillVec4(pixels, offs, particle.texCoordHead[0], particle.texCoordHead[1]);
```
magcius, line 826: you could pack scale and texCoordHead into one pixel and use 3 pixels per particle but who cares

src/WorldOfWarcraft/data.ts
```diff
@@ -465,6 +1016,11 @@ export class ModelData {
         mat4.mul(this.boneTransforms[i], this.boneTransforms[parentId], this.boneTransforms[i]);
       }
     }
+
+    // if (this.fileId === 197012) debugger;
```
magcius, line 1020: same

src/WorldOfWarcraft/data.ts
wgreenberg: lmao i had a feeling i was reinventing a wheel

src/WorldOfWarcraft/data.ts
wgreenberg: nice

src/WorldOfWarcraft/data.ts
```diff
@@ -303,6 +304,532 @@ export class SkyboxData {
...
+  private updateBuffer: Float32Array;
+  private particlesToEmit = 0.0;
+  private dataTexture: GfxTexture | undefined;
+  private textureColMask: number;
+  private textureColBits: number;
+  public particleType: number;
+  public spline?: BezierSpline;
+
+  constructor(public index: number, public emitter: WowM2ParticleEmitter, private model: ModelData, public txac: number) {
+    this.updateBuffer = new Float32Array(11);
```
magcius, line 571: One more thing, here you use `new Float32Array(32)` but down below in the gravity code, you reference `this.updateBuffer[11]` and `this.updateBuffer[12]`. Guessing this is wrong?

src/WorldOfWarcraft/data.ts
```diff
@@ -303,6 +304,532 @@ export class SkyboxData {
...
+    this.wind = convertWowVec3(emitter.wind_vector);
+    this.position = convertWowVec3(emitter.position);
```
magcius, line 573: btw, idle thought, but since you found a way of passing Float32Array to Rust, I wonder if you could do the same here so you wouldn't need a convertWowVec3 thing?

src/WorldOfWarcraft/data.ts
```diff
@@ -303,6 +304,532 @@ export class SkyboxData {
...
+        particle.color[0] = this.updateBuffer[0] / 255.0;
+        particle.color[1] = this.updateBuffer[1] / 255.0;
+        particle.color[2] = this.updateBuffer[2] / 255.0;
+        particle.color[3] = this.updateBuffer[3];
+        particle.scale[0] = this.updateBuffer[4];
+        particle.scale[1] = this.updateBuffer[5];
+        const cellHead = this.updateBuffer[6];
+        this.extractTexCoords(particle.texCoordHead, cellHead);
+        const cellTail = this.updateBuffer[6];
+        this.extractTexCoords(particle.texCoordTail, cellTail);
```
magcius, line 793: are these both supposed to be using updateBuffer[6] ?

src/WorldOfWarcraft/data.ts
wgreenberg: indeed it is!

src/WorldOfWarcraft/data.ts
wgreenberg: it's true, though it'd require a ton of boilerplate to store the deku-marshalled rust values into a buffer -- i also think syntactically it's convenient to have individual vec3s to reference, rather than juggling indices everywhere.

probably worth doing if this ends up becoming a bottleneck, though

src/WorldOfWarcraft/data.ts
wgreenberg: not at all

src/WorldOfWarcraft/data.ts
magcius: yeah, all good, just a thought

PR #704, opened 2024-09-21
wow: use clipped frustums to cull exterior

wgreenberg:
When we clip a frustum through a WMO group's portal, use that clipped
frustum to cull things outside. Also, perform the group membership check
in rust, which is both more convenient and faster.

Fixes a couple bugs also

### DIFF
--- a/rust/src/geometry.rs
+++ b/rust/src/geometry.rs
@@ -1,6 +1,6 @@
 use core::f32;
 
-use nalgebra_glm::{make_mat4, make_vec3, triangle_normal, vec3, vec4, Mat4, Vec3};
+use nalgebra_glm::{make_mat4, make_vec3, triangle_normal, vec2, vec3, vec4, Mat4, Vec3, Vec2};
 use wasm_bindgen::prelude::*;
 
 #[derive(Default, Debug, Clone)]
@@ -24,9 +24,8 @@ impl Plane {
         self.normal.dot(p) + self.d
     }
 
-    pub fn intersect_line(&self, p: &Vec3, dir: &Vec3) -> Vec3 {
-        let t = -(self.normal.dot(p) + self.d) / self.normal.dot(dir);
-        p + t * dir
+    pub fn intersect_line(&self, p: &Vec3, dir: &Vec3) -> f32 {
+        -(self.normal.dot(p) + self.d) / self.normal.dot(dir)
     }
 
     pub fn transform(&mut self, inv_transpose_mat: &Mat4) {
@@ -36,6 +35,23 @@ impl Plane {
         self.normal.y = transformed.y;
         self.normal.z = transformed.z;
         self.d = transformed.w;
+    }
+
+    pub fn major_axis(&self) -> Axis {
+        let mut max_axis = f32::NEG_INFINITY;
+        let mut axis = 3;
+        for i in 0..3 {
+            if self.normal[i].abs() > max_axis {
+                max_axis = self.normal[i].abs();
+                axis = i as u8;
+            }
+        }
+        match axis {
+            0 => Axis::X,
+            1 => Axis::Y,
+            2 => Axis::Z,
+            _ => unreachable!(),
+        }
     }
 }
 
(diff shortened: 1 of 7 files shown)


src/SourceEngine/BSPFile.ts
```diff
@@ -1568,8 +1568,8 @@ export class BSPFile {
             const surfaceFogVolumeID = facelist.getUint16(idx + 0x0E, true);
 
             const area = facelist.getFloat32(idx + 0x18, true);
-            const m_LightmapTextureMinsInLuxels = nArray(2, (i) => facelist.getInt32(idx + 0x1C + i * 4, true));
-            const m_LightmapTextureSizeInLuxels = nArray(2, (i) => facelist.getUint32(idx + 0x24 + i * 4, true));
+            const m_LightmapTextureMinsInLuxels = nArray(2, (i) => facelist!.getInt32(idx + 0x1C + i * 4, true));
+            const m_LightmapTextureSizeInLuxels = nArray(2, (i) => facelist!.getUint32(idx + 0x24 + i * 4, true));
```
magcius, line 1572: Interesting that I'm not hitting this. I wonder if you're using an older TypeScript version through Zed or something? If you're using VS Code, I pushed a fix to try and make it use the local repository version in 86773671fcda1e60df932b4243dc8fcaffa44a50.

src/WorldOfWarcraft/data.ts
```diff
@@ -929,6 +930,9 @@ export class WmoData {
 
         for (const fileId of this.wmo.group_file_ids) {
             this.groupDescriptors.push(this.wmo.get_group_descriptor(fileId));
+            if (this.groupDescriptors[this.groupDescriptors.length - 1].antiportal) {
+                console.log('antiportal detected!!!', fileId);
```
magcius, line 934: you might want to make this `console.warn` or add a `debugger;` statement after this

src/WorldOfWarcraft/scenes.ts
```diff
@@ -261,7 +261,12 @@ export class MapArray<K, V> {
     }
 }
 
-enum CullWmoResult {
+interface CullWmoResult {
+    cameraState: CullCameraState,
+    frustums?: ConvexHull[],
```
magcius, line 266: semicolons not commas in TypeScript

rust/src/geometry.rs
```diff
@@ -24,9 +24,8 @@ impl Plane {
         self.normal.dot(p) + self.d
     }
 
-    pub fn intersect_line(&self, p: &Vec3, dir: &Vec3) -> Vec3 {
-        let t = -(self.normal.dot(p) + self.d) / self.normal.dot(dir);
-        p + t * dir
+    pub fn intersect_line(&self, p: &Vec3, dir: &Vec3) -> f32 {
```
magcius, line 27: I might write a comment to the effect of

`// Returns the distance t of the intersection of this plane with the line p + t*dir`

rust/src/geometry.rs
```diff
@@ -113,6 +133,14 @@ impl AABB {
             || p.z < self.min.z
             || p.z > self.max.z)
     }
+
+    pub fn contains_point_ignore_max_z(&self, p: &Vec3) -> bool {
+        !(p.x < self.min.x
```
magcius, line 138: nitpick; This check is specific enough that I might move it out of the public API, but I don't care too much

rust/src/wow/wmo.rs
```diff
@@ -729,10 +777,8 @@ pub struct BspTreeResult {
 
 fn neg_z_line_intersection(
     p: &nalgebra_glm::Vec3,
-    tri: (nalgebra_glm::Vec3, nalgebra_glm::Vec3, nalgebra_glm::Vec3),
+    (vertex0, vertex1, vertex2): (nalgebra_glm::Vec3, nalgebra_glm::Vec3, nalgebra_glm::Vec3),
```
magcius, line 780: nitpick: I prefer the term position rather than vertex, as a vertex can have more data than just its positions

rust/src/wow/wmo.rs
```diff
@@ -1258,12 +1317,15 @@ impl WmoGroupFlags {
...
+            unreachable: x & 0x80 > 0,
             show_exterior_sky: x & 0x100 > 0,
             has_lights: x & 0x200 > 0,
             has_doodads: x & 0x800 > 0,
             has_water: x & 0x1000 > 0,
             interior: x & 0x2000 > 0,
+            always_draw: x & 0x10000 > 0,
             show_skybox: x & 0x40000 > 0,
+            unk_skip_membership_check: x & 0x400000 > 0,
             water_is_ocean: x & 0x80000 > 0,
```
magcius, line 1329: mind keeping these sorted by value? 0x400000 is more than 0x80000

PR #712, opened 2024-10-22
Implement d_wood (bushes) for Wind Waker

themikelester:
Screenshot 2024-10-21 at 8 39 13 PM

Full rendering support, as well as idle and cut animation support. This is picking up from a changelist I stashed 5 years ago 😂


src/ZeldaWindWaker/SComponent.ts
```diff
@@ -129,6 +129,24 @@ export function cM__Deg2Short(v: number): number {
     return cM__Rad2Short(v * MathConstants.DEG_TO_RAD);
 }
 
+// Move `value` closer to `target` by the amount specified in `step`. 
+// If `value` would cross `target`, it is set to `target`. Returns 1 if value has reached the target, 0 otherwise.
+export function cLib_chaseS(value: { x: number }, target: number, step: number): number {
```
magcius, line 134: I think this is equivalent to `cLib_chaseF`? Mind using that instead? Feel free to move the documentation over.

src/ZeldaWindWaker/d_wood.ts
```diff
@@ -0,0 +1,890 @@
...
+// Types
+//-----------------------------------------
+const enum UnitFlags {
+    Inactive = 0,
+    Active = 1 << 0,
+    IsFrustumCulled = 1 << 1,
+    IsCut = 1 << 2,
+}
+
+enum AnimMode_e {
```
magcius, line 36: `const enum` please.

src/ZeldaWindWaker/d_wood.ts
```diff
@@ -0,0 +1,890 @@
...
+    Norm = 4,     // Idle animation
+    ToNorm = 5,   // Blend back to the normal animation
+
+    _Max
+};
+
+//-----------------------------------------
+// Globals
+//-----------------------------------------
+let globals: dGlobals;
```
magcius, line 50: No. Pass `dGlobals` everywhere that's required.

src/ZeldaWindWaker/d_wood.ts
```diff
@@ -0,0 +1,890 @@
...
+    constructor(lGlobals: dGlobals) {
+        globals = lGlobals;
+        this._mModel = new WoodModel(lGlobals);
+
+        for (let i = 0; i < 8; i++) {
+            this.mAnm[i].mode_norm_init();
+        }
+    }
+
+    destroy(device: GfxDevice) {
```
magcius, line 662: Wind Waker uses `public`/`private`/`protected` attributes.

src/ZeldaWindWaker/d_wood.ts
```diff
@@ -0,0 +1,890 @@
...
+        const l_color = globals.findExtraSymbolData('d_wood.o', 'l_color__Q25dWood20@unnamed@d_wood_cpp@');
+        const l_pos = globals.findExtraSymbolData('d_wood.o', 'l_pos__Q25dWood20@unnamed@d_wood_cpp@');
+        const l_Txa_swood_bTEX = globals.findExtraSymbolData('d_wood.o', 'l_Txa_swood_bTEX__Q25dWood20@unnamed@d_wood_cpp@');
+        const l_shadowVtxDescList = globals.findExtraSymbolData('d_wood.o', 'l_shadowVtxDescList$5139');
+        const l_shadowVtxAttrFmtList = globals.findExtraSymbolData('d_wood.o', 'l_shadowVtxAttrFmtList$5140');
+        const l_vtxDescList = globals.findExtraSymbolData('d_wood.o', 'l_vtxDescList$5156');
+        const l_vtxAttrFmtList = globals.findExtraSymbolData('d_wood.o', 'l_vtxAttrFmtList$5157');
+
+        // @HACK: The tex coord array is being read as all zero. Hardcode it.
+        const l_shadowTexCoord = new ArrayBufferSlice(new Uint8Array([0, 0, 1, 0, 1, 1, 0, 1]).buffer);
```
magcius, line 252: these aren't all zeroes though?

src/ZeldaWindWaker/d_wood.ts
```diff
@@ -0,0 +1,890 @@
...
+        while (this.mRootUnit) {
+            let unit = this.mRootUnit;
+            this.mRootUnit = unit.mNextUnit!;
+            unit.clear();
+        }
+    }
+}
+
+class Anm_c {
+    /* 0x00 */ mModelMtx: mat4 = mat4.create();
```
magcius, line 350: Mind removing the offset comments, the m prefix, and adding public/private?

src/ZeldaWindWaker/d_wood.ts
```diff
@@ -0,0 +1,890 @@
...
+            case AnimMode_e.PushBack: return this.mode_push_back(packet);
+            case AnimMode_e.Fan: return this.mode_fan(packet);
+            case AnimMode_e.Norm: return this.mode_norm(packet);
+            case AnimMode_e.ToNorm: return this.mode_to_norm(packet);
+            default: return;
+        }
+    }
+
+    public copy_angamp(anm: Anm_c): void {
+
```
magcius, line 384: Leave TODO comments if you haven't implemented it, or remove the method if it's unnecessary for the noclip implementation.

src/ZeldaWindWaker/d_wood.ts
```diff
@@ -0,0 +1,890 @@
...
+                phase = 1;
+            } else {
+                phase = 2;
+            }
+        }
+
+        let fVar1 = 0.0;
+        let fVar6 = fVar1;
+        for (let i = 0; i < 2; i++) {
+            const animAttr = l_animAttrs[phase][i];
```
magcius, line 496: Mind documenting these if you can?

src/ZeldaWindWaker/d_wood.ts
```diff
@@ -0,0 +1,890 @@
...
+            vec3.set(scratchVec3a, 0, 1, 0);
+        }
+
+        const normal = scratchVec3a;
+        const right = vec3.set(scratchVec3c, 1, 0, 0);
+        const forward = vec3.cross(scratchVec3d, normal, right);
+        vec3.cross(right, normal, forward);
+
+        // Get the normal from the raycast, rotate shadow to match surface
+        this.mShadowModelMtx[0] = right[0];
```
magcius, line 566: setMatrixAxis and setMatrixTranslation might be helpful here.

src/ZeldaWindWaker/d_wood.ts
```diff
@@ -0,0 +1,890 @@
...
+    /**
+     * Compute modelView matrices for the body, trunk, and drop shadow
+     * @param anim 
+     */
+    public set_mtx(anims: Anm_c[]): void {
+        mDoMtx_copy(anims[this.mAnmIdx].mModelMtx, scratchMat4a);
+        scratchMat4a[12] += this.mPos[0];
+        scratchMat4a[13] += this.mPos[1];
+        scratchMat4a[14] += this.mPos[2];
+        mat4.mul(this.mModelViewMtx, globals.camera.viewMatrix, scratchMat4a);
```
magcius, line 595: Probably easier to construct the MVM at prepareToRender time and just add animation/model matrices in here.

src/ZeldaWindWaker/d_wood.ts
```diff
@@ -0,0 +1,890 @@
...
+
+        // TODO: Add to the Render List
+    }
+
+    public draw(globals: dGlobals, renderInstManager: GfxRenderInstManager, viewerInput: ViewerRenderInput): void {
+        const worldToView = viewerInput.camera.viewMatrix;
+        const worldCamPos = mat4.getTranslation(scratchVec3b, viewerInput.camera.worldMatrix);
+
+        // Render to the XLU BG display list (after the bg terrain). We want to render late since we are alpha tested.
+        renderInstManager.setCurrentList(globals.dlst.bg[1]);
```
magcius, line 804: Alpha tested shouldn't matter? Do you have depth write on or not.

src/ZeldaWindWaker/d_wood.ts
```diff
@@ -0,0 +1,890 @@
...
+        displayListRegistersInitGX(matRegisters);
+        displayListRegistersRun(matRegisters, l_matDL);
+
+        const material = parseMaterial(matRegisters, 'd_tree::l_matDL');
+        material.alphaTest.op = GX.AlphaOp.OR;
+        material.alphaTest.compareA = GX.CompareType.GREATER;
+        material.alphaTest.compareB = GX.CompareType.GREATER;
+        material.alphaTest.referenceA = kAlphaCutoff;
+        material.alphaTest.referenceB = kAlphaCutoff;
+        material.hasDynamicAlphaTest = true;
```
magcius, line 287: Any reason you turn on dynamic alpha test here?

src/ZeldaWindWaker/tools/zww_extractor.ts
```diff
@@ -332,7 +346,9 @@ function extractExtra(binaries: Binary[]) {
     };
 
     const data = BYML.write(crg1, BYML.FileType.CRG1);
+    mkdirSync(pathBaseOut, { recursive: true });
     writeFileSync(`${pathBaseOut}/extra.crg1_arc`, Buffer.from(data));
+    cpSync(`${pathBaseIn}/res`, pathBaseOut, {recursive: true})
```
magcius, line 351: I might want to be a bit more discerning here, but sure.

src/ZeldaWindWaker/d_wood.ts
themikelester: Why is that? Isn't `globals` in this case just a reference to the same object?

src/ZeldaWindWaker/tools/zww_extractor.ts
themikelester: Oh my bad, I did not mean to include this change in this PR

src/ZeldaWindWaker/d_wood.ts
themikelester: Ah this is an old (5 years!) comment. Must have been an issue back then. I'll clean this up...

src/ZeldaWindWaker/d_wood.ts
themikelester: Will do!

src/ZeldaWindWaker/d_wood.ts
themikelester: Definitely!

src/ZeldaWindWaker/SComponent.ts
themikelester: Ah yea that's an interesting one. It _is_ cLib_chaseS in game (since it's chasing s16s) but in JS land I guess that doesn't much matter. I'll add cLib_chaseF too.

src/ZeldaWindWaker/d_wood.ts
themikelester: Nice!

src/ZeldaWindWaker/d_wood.ts
themikelester: Ah I'm just matching what the cpp was doing, so I'm more confident I don't miss anything. Should I move it around? Got a preference?

src/ZeldaWindWaker/d_wood.ts
themikelester: Yea it's related to the cut down animation (which I've been playing with locally). This is the logic they use in the game. When cut, they immediately start dropping the alpha to fade the bush out. If alpha is less than 1.0, they disable alpha testing and set the reference alphas to 0 dynamically.

src/ZeldaWindWaker/d_wood.ts
magcius: Yes.

src/ZeldaWindWaker/d_wood.ts
magcius: I'm saying that alpha tested / depth write / blend off should be order-independent, and act just like opaque geometry.

src/ZeldaWindWaker/d_wood.ts
magcius: So it effectively switches to alpha blending in this case, OK. I might suggest that you create two GX materials for clarity, but this is fine.

src/ZeldaWindWaker/d_wood.ts
magcius: I don't like module-level globals for many reasons, but in this case, if you switch to a different game, it won't clear the contents of `globals`, effectively leaking a whole copy of Wind Waker's engine and resources. You could fix that by having a destructor that clears it out, but passing it around is cleaner, and what I do basically everywhere else.

src/ZeldaWindWaker/d_wood.ts
```diff
@@ -0,0 +1,867 @@
...
+    public copy_angamp(anm: Anm_c): void {
+
+    }
+
+    // Animations are assigned from the Packet to specific Wood instances (Bushes) when a new animation starts
+    // Each animation mode has an mode_*_init() function which is called when the animation is started
+    // The mode_*() function is called to update the animation each frame, until finished
+
+    public mode_cut_init(anm: Anm_c, targetAngle: number): void {
+        for (let i = 0; i < 2; i++) {
```
magcius, line 389: This function is never called? Also, you never use `anm`.

src/ZeldaWindWaker/m_do_mtx.ts
```diff
@@ -39,6 +39,10 @@ export function mDoMtx_XYZrotM(dst: mat4, v: vec3): void {
     mat4.rotateX(dst, dst, cM__Short2Rad(v[0]));
 }
 
+export function mDoMtx_copy(src: ReadonlyMat4, dst: mat4): void {
```
magcius, line 42: I tend to just write this inline, but sure.

src/ZeldaWindWaker/d_wood.ts
```diff
@@ -0,0 +1,867 @@
...
+}
+
+//-----------------------------------------
+// Classes
+//-----------------------------------------
+/**
+ * A linked list of Units, so that they can be traversed room by room
+ */
+class Room_c {
+    mRootUnit: Unit_c;
```
magcius, line 329: You could make this as a linked list, but in Grass, we turn the entire Room_c equivalent into just an array. I think you could replace this with `private rooms: Unit_c[][] = nArray(64, () => []);` on Packet_c.

magcius (commented):
Looking way better, thank you! Almost ready to merge, just a few nits, and I need to upload a new extra arc.

src/ZeldaWindWaker/d_wood.ts
```diff
@@ -0,0 +1,867 @@
...
+        }
+
+        const normal = scratchVec3a;
+        const right = vec3.set(scratchVec3c, 1, 0, 0);
+        const forward = vec3.cross(scratchVec3d, normal, right);
+        vec3.cross(right, normal, forward);
+
+        // Get the normal from the raycast, rotate shadow to match surface
+        setMatrixAxis(this.mShadowModelMtx, right, normal, forward);
+        mat4.transpose(this.mShadowModelMtx, this.mShadowModelMtx);
```
magcius, line 556: This transpose shouldn't be needed anymore.

src/ZeldaWindWaker/d_wood.ts
themikelester: The alpha test failure seems to not actually do a discard. It seems like depth is being written along the edges where the alpha test fails. Here's a screenshot when the bushes draw before the bg terrain: 
Screenshot 2024-10-23 at 9 52 13 PM
Actually as I'm looking at this it seems clear that blending is enabled. But I'm 90% sure this is how the game does it. You can see the same artifacts (ground color comes through along leaf edges) on the gamecube.

src/ZeldaWindWaker/d_wood.ts
themikelester: Ah yes, this matches the cpp. I guess there's no point in keeping those leftovers around

themikelester: As for the function, it's part of a bigger change that I'm working on. I'd like to keep it in, if that's alright.

src/ZeldaWindWaker/d_wood.ts
themikelester: Yea this is just mimicking the cpp again. But now that I've got my head around how the cpp is written it's clear there's no harm in skipping the list.

themikelester:
Yea thanks for the thorough review! It's been so long, I'd like to match your vision as best I can. Rather have all the nits up front 😁

src/ZeldaWindWaker/d_wood.ts
```diff
@@ -0,0 +1,826 @@
...
+    mAmpY: number[] = [0, 0];
+    mAmpX: number[] = [0, 0];
+
+    mNextAnimIdx: number; // Corresponds to the index in Packet_c::mAnm;
+
+    mAlpha: number = 0xFF;
+
+    public play(packet: Packet_c): void {
+        switch (this.mMode) {
+            case AnimMode_e.Cut: return this.mode_cut(packet);
```
magcius, line 348: You're passing an extra argument here causing the build to fail.

src/ZeldaWindWaker/d_wood.ts
```diff
@@ -0,0 +1,826 @@
...
+            case AnimMode_e.PushInto: return this.mode_push_into(packet);
+            case AnimMode_e.PushBack: return this.mode_push_back(packet);
+            case AnimMode_e.Fan: return this.mode_fan(packet);
+            case AnimMode_e.Norm: return this.mode_norm(packet);
```
magcius, line 352: Same here.

PR #722, opened 2024-11-19
wow: fix ParticleEmitter texture leak

wgreenberg:


src/WorldOfWarcraft/data.ts
```diff
@@ -166,9 +166,10 @@ type WmoDefinitionKey = [number, number, number];
 
 export class WowCache {
     private sheepfile: Sheepfile;
-    private promiseCache = new Map<number, Promise<unknown>>();
+    private promiseCache = new Map<number, Promise<any>>();
```
magcius, line 169: This seems unintentional but OK.

PR #725, opened 2024-11-21
Use node's `createRequire()` to get the `require` function for `loadRustLib()`

themikelester:
`require` does not exist in the global namespace, at least on my machine. According to the node.js docs, it shouldn't. This uses a builtin node function to generate the `require` function for use within ES modules.

Without this change, I can't run tools such as zww_extractor.ts

See https://nodejs.org/api/esm.html#no-require-exports-or-moduleexports

### DIFF
--- a/src/rustlib.ts
+++ b/src/rustlib.ts
@@ -1,9 +1,12 @@
 
 import init, * as rust from '../rust/pkg/noclip_support';
+import { createRequire } from 'node:module';
 
 export { rust };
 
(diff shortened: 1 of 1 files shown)


themikelester:
Whoops. I didn't realize that `loadRustLib()` was used by more than just tools. This change won't work in browser environments. So the issue seems to be that we need to support using `require` in both the browser and node (for tools). I'll work on a better fix.

PR #726, opened 2024-11-22
Wind Waker: Allow users to refresh, save/load, or share cutscene scenes at a specific point in time

themikelester:
The current demo time and pause state are now added to the save state string. 

This allows users to refresh, save/load, or share demo scenes at a specific point in time. Since the pause state is also captured, if the demo was paused it will still be paused on load. This is also very useful for debugging. Pausing during certain actor animations will save the state, and then subsequent code changes will trigger refreshes which will load the demo back into the same state.

The big change here is that we need to a way to expose to the scene a way to play/pause noclip. I've done that by exposing the `UI.togglePlayPause()` method because of its simplicity, but there are several other alternatives.


src/viewer.ts
```diff
@@ -58,6 +58,7 @@ export interface SceneGfx {
     serializeSaveState?(dst: ArrayBuffer, offs: number): number;
     deserializeSaveState?(src: ArrayBuffer, offs: number, byteLength: number): number;
     onstatechanged?: () => void;
+    togglePlayPause?: (shouldBePlaying?: boolean) => void;
```
magcius, line 61: The better way to wire this in would be pass it in as part of SceneContext, like a `context.timeManager` class or something.

magcius (commented):
I'm not sure I see the utility of this? At one point I had the time and pause state in the save state, but removed it since people would get linked to a paused save state and end up confused.

themikelester:
For developing, it's super useful to be able to save/load from a specific time in a cutscene (and it needs to load as paused to be useful). 

For users, I could see utility in being able to share an interesting point in a cutscene. I wouldn't want to add time and pause state to _all_ save states, but it feels important for demos.

themikelester:
Oh wow I never knew about IS_DEVELOPMENT. Good thinking, that should work well for my purposes.

themikelester:
I've updated this PR so that the scene's time/pause state is not included in the savestate. If requested, a scene can save and load the time/pause state to localStorage. ZWW demo scenes are currently the only scenes that do this.

Changes for all scenes:
- Scenes can control their play/pause state via `SceneContext.SceneTime.togglePlayPause()`
- Scenes can optionally save/load/delete their time/pause state via `SceneContext.SceneTime`

Changes for ZWW demo scenes:
- The demo timeline is now synced to the scene's time. Setting the scene time before a demo is played will start the demo at the correct time.
- In development mode: When a demo is paused, it's time/pause state is saved to localStorage. Reloading the page will load the demo at the same time and paused. Resuming the demo (or switching scenes) clears this state and reloading will restart the demo.

magcius:
Sorry for the delay, I was on vacation for a week. I decided I wanted to do a bigger refactor of the scene loading and timekeeping code. Now it will save playing state, scene time, and scene time scale between refreshes.

themikelester:
Cool!!

PR #732, opened 2024-11-29
Wind Waker: Link now appears in cutscenes (Neat!!)

themikelester:
Screenshot 2024-11-28 at 5 00 01 PM
Screenshot 2024-11-28 at 5 01 47 PM

This is a pretty straightforward actor implementation in d_a.ts, but I wanted to put it up as a PR because it'll require uploading a new `extra.crg1_arc` because of Link's animation table. 

Most of the functionality is there, still a few things to clean up. Item equipping (mostly the sword), pupil animation, etc.


src/ZeldaWindWaker/d_a.ts
```diff
@@ -4955,6 +4961,661 @@ class d_a_npc_ls1 extends fopNpc_npc_c {
...
+        // Render an alpha mask in the shape of the eyes. Needs to render before Link so that it can depth test against
+        // the scene but not against his hair. The eyes will then draw with depth testing enabled, but will mask against
+        // this alpha tex. 
+        matInstA.setSortKeyLayer(GfxRendererLayer.OPAQUE + 4, false);
+        matInstA.setColorWriteEnabled(false);
+        matInstA.setAlphaWriteEnabled(true);
+
+        // @TODO: This material is marked as translucent in the original BMD. Noclip draws translucent shapes after all
+        //        opaque shapes, meaning that this renders AFTER Link, which defeats the purpose of modifying the sort
+        //        layer. Is there something wrong with noclip's translucency handling?
```
magcius, line 5205: The default mDoExt_modelEntryDL draws opaque shapes to drawListSet[0], and translucent shapes to drawListSet[1]. The only thing that says that one is after the other is the code that draws the lists in Main.ts. A GfxRenderInstList is sort of equivalent to a J3DDrawBuffer.

I'm guessing the original game has some very specific handling for this in the draw function that would take some work to replicate in noclip. My Link actor hacks were done on the quick for video work.

src/ZeldaWindWaker/Main.ts
```diff
@@ -1014,11 +1020,15 @@ class DemoDesc extends SceneDesc implements Viewer.SceneDesc {
...
                 })
-
-                await globals.modelCache.waitForLoad();
             }
         }
 
+        await globals.modelCache.waitForLoad();
+        
+        if(!fopAcM_searchFromName(globals, 'Link', 0, 0)) {
+            fopAcM_create(globals.frameworkGlobals, dProcName_e.d_a_py_lk, 0, [-25414,3501,219], globals.mStayNo, null, null, 0xFF, -1);
```
magcius, line 1029: Where does this come from and why? Seems wrong.

src/ZeldaWindWaker/d_a.ts
```diff
@@ -4955,6 +4961,661 @@ class d_a_npc_ls1 extends fopNpc_npc_c {
...
+    private demoProcInitFuncTable = new Map<LinkDemoMode, () => boolean>();
+    private proc: (globals: dGlobals) => void;
+
+    private model: J3DModelInstance;
+    private modelSwordHilt: J3DModelInstance;
+    private modelKatsura: J3DModelInstance; // Wig. To replace the hat when wearing casual clothes.
+
+    private demoMode: number = LinkDemoMode.None;
+    private demoClampToGround = true;
+    private gdnChk = new dBgS_GndChk()
```
magcius, line 5041: gdnChk? Is that a typo?

src/ZeldaWindWaker/d_a.ts
```diff
@@ -4955,6 +4961,661 @@ class d_a_npc_ls1 extends fopNpc_npc_c {
...
+            this.changeDemoProc(globals);
+        }
+
+        // Step our animations forward
+        this.anmBck.play(deltaTimeFrames);
+        this.anmBtp.play(deltaTimeFrames);
+        this.anmBtk.play(deltaTimeFrames);
+
+        // Run the current custom update process (Walk, Idle, Swim, etc)
+        if (this.proc) this.proc(globals);
```
magcius, line 5097: I'm surprised this passes the correct this object. In general you would need a call/apply, bind on set, or use arrow functions for the definitions of proc methods. See the modeProcInit/modeProcExec helpers I have for this.

src/ZeldaWindWaker/d_a.ts
```diff
@@ -4955,6 +4961,661 @@ class d_a_npc_ls1 extends fopNpc_npc_c {
...
+            this.modelSwordHilt.calcAnim();
+        }
+    }
+
+    private procWait_init(globals: dGlobals) {
+        this.setSingleMoveAnime(globals, LkAnim.WAITS);
+    }
+
+    private loadAnmTable(globals: dGlobals) {
+        const anmDataView = globals.findExtraSymbolData(`d_a_player_main.o`, `mAnmDataTable__9daPy_lk_c`).createDataView();
```
magcius, line 5604: I wouldn't be opposed if you just embedded this table directly into the actor source code.

magcius (changes requested):

src/ZeldaWindWaker/Main.ts
themikelester: Good catch, the position is a leftover (it's his position on Outset during the title cutscene, but it depends on that demos offset). I'll just spawn him at the origin...

src/ZeldaWindWaker/d_a.ts
themikelester: You've never head of a grouden check?! Good catch 😅

src/ZeldaWindWaker/d_a.ts
themikelester: Hmm I'm surprised it works too. I thought I would have called `.bind()` when I set the proc, but I must have slipped it. I'll check out `modeProcInit`...

src/ZeldaWindWaker/d_a.ts
themikelester: I started that way, but I think I'm going to be adding more animations soon and the data for each is 5 properties so it gets verbose quickly. Plus its fun to be able to call any Link anim by name just to try em all out. 

I'll leave it up to you.

src/ZeldaWindWaker/d_a.ts
themikelester: Ah I see, that's very clear. I've modified the comment.

src/ZeldaWindWaker/Main.ts
magcius: Won't this spawn Link on all maps, including those without cutscenes or Link in those cutscenes? I'd prefer if the Link actor was spawned when a cutscene is started, if possible.

src/ZeldaWindWaker/Main.ts
```diff
@@ -1014,11 +1020,15 @@ class DemoDesc extends SceneDesc implements Viewer.SceneDesc {
...
-
-                await globals.modelCache.waitForLoad();
             }
         }
 
+        await globals.modelCache.waitForLoad();
+        
+        if(!fopAcM_searchFromName(globals, 'Link', 0, 0)) {
+            fopAcM_create(globals.frameworkGlobals, dProcName_e.d_a_py_lk, 0, null, globals.mStayNo, null, null, 0xFF, -1);
+        }
```
magcius, line 1030: As mentioned before, I don't want a Link spawning in every map. Spawn it in the cutscene system if you must.

src/ZeldaWindWaker/Main.ts
```diff
@@ -1000,6 +1000,12 @@ class DemoDesc extends SceneDesc implements Viewer.SceneDesc {
 
         // TODO: Don't render until the camera has been placed for this demo. The cuts are jarring.
 
+        // Most cutscenes expect the Link actor to be loaded
+        this.globals.modelCache.fetchObjectData('Link');
+        this.globals.modelCache.fetchObjectData('LkD00');
+        this.globals.modelCache.fetchObjectData('LkD01');
+        this.globals.modelCache.fetchObjectData('LkAnm');
+
```
magcius, line 1008: Can you load these in the Link actor? I'm not sure why this is necessary here.

src/ZeldaWindWaker/d_a.ts
```diff
@@ -4955,6 +4961,673 @@ class d_a_npc_ls1 extends fopNpc_npc_c {
...
+     * Get the angle of the ground based when facing a specific direction
+     * @param dir the s16 angle which the actor is facing
+     */
+    private getGroundAngle(globals: dGlobals, dir: number) {
+        const gndPlane = globals.scnPlay.bgS.GetTriPla(this.gndChk.polyInfo.bgIdx, this.gndChk.polyInfo.triIdx);
+        const norm = gndPlane.n;
+    
+        if (gndPlane && norm[1] >= 0.5) {
+            const slopeDir = cM_atan2s(norm[0], norm[2]);
+            const slopeGrade = Math.sqrt(norm[0] * norm[0] + norm[2] * norm[2]);
```
magcius, line 5434:
```suggestion
            const slopeGrade = Math.hypot(norm[0], norm[2]);
```

src/ZeldaWindWaker/d_a.ts
```diff
@@ -4955,6 +4961,673 @@ class d_a_npc_ls1 extends fopNpc_npc_c {
...
+                } else {
+                    this.handStyleRight = LkHandStyle.Idle;
+                }
+            }
+        }
+
+        if (anmBckId == 0xFFFF || this.anmBckId == anmBckId) {
+            if (demoActor.flags & EDemoActorFlags.HasFrame) {
+                this.anmBck.frameCtrl.currentTimeInFrames = anmFrame;
+                this.anmBtp.frameCtrl.currentTimeInFrames = anmFrame;
```
magcius, line 5549:
```suggestion
                this.anmBck.setFrame(anmFrame);
                this.anmBtp.setFrame(anmFrame);
```

src/ZeldaWindWaker/d_a.ts
```diff
@@ -4955,6 +4961,673 @@ class d_a_npc_ls1 extends fopNpc_npc_c {
...
+                demoActor.animFrameMax = this.anmBck.frameCtrl.endFrame;
+            }
+        } else {
+            // TODO: How should LkD00 arc be loaded?
+            const bck = globals.resCtrl.getObjectIDRes(ResType.Bck, 'LkD00', anmBckId);
+            this.anmBck.init(this.model.modelData, bck, true, bck.loopMode, 1.0, 0, bck.duration);
+            this.anmBck.frameCtrl.currentTimeInFrames = anmFrame;
```
magcius, line 5556:
```suggestion
            this.anmBck.setFrame(anmFrame);
```

src/ZeldaWindWaker/d_a.ts
```diff
@@ -4955,6 +4961,673 @@ class d_a_npc_ls1 extends fopNpc_npc_c {
...
+        // noclip modification:
+        this.setSingleMoveAnime(globals, LkAnim.WAITS);
+
+        return cPhs__Status.Next;
+    }
+
+    override execute(globals: dGlobals, deltaTimeFrames: number): void {
+        // Update the current proc based on demo data
+        this.setDemoData(globals);
+        if (this.demoMode != 5) {
```
magcius, line 5094:
```suggestion
        if (this.demoMode !== LinkDemoMode.WaitTurn) {
```

src/ZeldaWindWaker/d_a.ts
```diff
@@ -4955,6 +4961,673 @@ class d_a_npc_ls1 extends fopNpc_npc_c {
...
+        moveVel *= Math.cos(cM_s2rad(groundAngle));        
+
+        // ... Reduce velocity even more for ascending slopes
+        if (groundAngle < 0) {
+            moveVel = moveVel * 0.85;
+        }
+
+        // Update actor vel and position
+        this.vel[0] = moveVel * Math.sin(cM_s2rad(this.rot[1]));
+        this.vel[1] = moveVel * Math.cos(cM_s2rad(this.rot[1]));
```
magcius, line 5415: Should this not be `this.vel[2]`? Y axis is usually up.

src/ZeldaWindWaker/d_a.ts
```diff
@@ -4955,6 +4961,673 @@ class d_a_npc_ls1 extends fopNpc_npc_c {
...
+        this.equippedItemModel = null;
+    }
+
+    private setItemModel() {
+        if(!this.equippedItemModel) {
+            return;
+        }
+
+        const handLJointMtx = this.model.shapeInstanceState.jointToWorldMatrixArray[0x08];
+        const handRJointMtx = this.model.shapeInstanceState.jointToWorldMatrixArray[0x0D];
```
magcius, line 5604: Sometimes you have constants for these or search by name, other times you don't?

src/ZeldaWindWaker/d_a.ts
```diff
@@ -4955,6 +4961,673 @@ class d_a_npc_ls1 extends fopNpc_npc_c {
...
+        this.posMove(globals);
+
+        // Evaluate for collisions, clamp to ground
+        this.gndChk.Reset();
+        vec3.scaleAndAdd(this.gndChk.pos, this.pos, Vec3UnitY, 30.1);
+        const groundHeight = globals.scnPlay.bgS.GroundCross(this.gndChk);
+        this.autoGroundHit();
+
+        // If we're pulling position directly from the JStudio tool, ignore collisions and animation root motion
+        if (this.curMode == d_a_py_lk_mode.tool) {
```
magcius, line 5117: Please remember to use `===` and `!==` for comparisons in TypeScript. This goes for every comparison in this PR.

src/ZeldaWindWaker/d_a.ts
```diff
@@ -4955,6 +4961,673 @@ class d_a_npc_ls1 extends fopNpc_npc_c {
     }
 }
 
+enum LkAnim {
```
magcius, line 4964: If you are not using the stringification feature of enums, please mark them as `const enum`.

magcius (changes requested):

src/ZeldaWindWaker/Main.ts
themikelester: Ah that is what's happening now. This code only runs in the demo scenes (not regular scenes), all of which require that the Link actor be spawned before the demo is started.

src/ZeldaWindWaker/Main.ts
themikelester: Yes, totally. I think that's a better place.

src/ZeldaWindWaker/d_a.ts
themikelester: Good to know!

src/ZeldaWindWaker/d_a.ts
themikelester: Fixed

src/ZeldaWindWaker/d_a.ts
themikelester: Fixed

src/ZeldaWindWaker/d_a.ts
magcius: hm, I thought mDoExt_baseAnm had getFrame/setFrame on them directly, but that's just a TP feature. We should probably add those as well.

src/ZeldaWindWaker/d_a.ts
themikelester: Wow! Good catch. This code is currently only used for a small part of one demo (at least in the set of demos that I've been iterating on) so I hadn't noticed. But it does fix his positioning in that scene.

src/ZeldaWindWaker/d_a.ts
themikelester: Right! That's a good reminder.

src/ZeldaWindWaker/Main.ts
magcius: Ah, I missed that this was in DemoDesc.

src/ZeldaWindWaker/d_a.ts
themikelester: Fixed!

src/ZeldaWindWaker/d_a.ts
themikelester: Went ahead and just made an LkJoint enum (filled only with the ones currently used). I think it's clearer this way.

magcius (approved):

PR #735, opened 2024-12-12
J2D: Implement JSystem 2D rendering library, use it display Wind Waker place names 

themikelester:
Screenshot 2024-12-12 at 4 28 38 PM

- Added `src/Common/JSYSTEM/J2D.ts`
- Added new automatic file type handling, `BLO` (Binary Layout). 
- Added `fopMsgM` functionality to `framework.ts`
- Implemented place names handing in `src/ZeldaWindWaker/d_place_name.ts`

BLO files contain layout data for 2D elements. It is made up of a hierarchy of J2DPanes (essentially just a bounding box) and J2Pictures. Child elements inherit the transform from their parents. The entire hierarchy contained in a single root J2DScreen object. I followed the pattern of J3D and split up the loading functions from the runtime objects. 

J2D and BLO files can also be used to render message boxes during demos. I'm hoping that others will find uses for them as well. 

As for place names: currently, the only ones that will display are Outset Island and The Forsaken Fortress during their intro cutscenes.

### DIFF
--- a/src/Common/JSYSTEM/J2D.ts
+++ b/src/Common/JSYSTEM/J2D.ts
@@ -5,7 +5,7 @@ import { Color, colorNewFromRGBA8 } from "../../Color.js";
 import { GfxRenderInst, GfxRenderInstManager } from "../../gfx/render/GfxRenderInstManager.js";
 import * as GX_Material from '../../gx/gx_material.js';
 import * as GX from '../../gx/gx_enum.js';
-import { DrawParams, fillSceneParamsData, GXMaterialHelperGfx, MaterialParams, SceneParams, ub_SceneParamsBufferSize } from "../../gx/gx_render.js";
+import { ColorKind, DrawParams, fillSceneParamsData, GXMaterialHelperGfx, MaterialParams, SceneParams, ub_SceneParamsBufferSize } from "../../gx/gx_render.js";
 import { GfxClipSpaceNearZ, GfxDevice } from "../../gfx/platform/GfxPlatform.js";
 import { computeModelMatrixT, projectionMatrixForCuboid } from "../../MathHelpers.js";
 import { projectionMatrixConvertClipSpaceNearZ } from "../../gfx/helpers/ProjectionHelpers.js";
@@ -152,7 +152,7 @@ function readPAN1Chunk(buffer: ArrayBufferSlice, parent: PAN1 | null): PAN1 {
 
     let rot = 0;
     let basePos = 0;
-    let alpha = 0;
+    let alpha = 0xFF;
     let inheritAlpha = true;
 
     if(dataCount >= 7) { rot = view.getUint16(offset); offset += 2; }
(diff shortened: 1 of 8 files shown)


src/Common/JSYSTEM/J2D.ts
```diff
@@ -0,0 +1,460 @@
+import ArrayBufferSlice from "../../ArrayBufferSlice.js";
```
magcius, line 1: Something worth noting is that J2D has two versions. The version in Wind Waker is v1. v2 is found in Twilight Princess (and possibly other games like Pikmin 2?)

So you might you want to rename this to J2Dv1.ts or something.

src/Common/JSYSTEM/J2D.ts
```diff
@@ -0,0 +1,460 @@
...
+                default: console.warn('Unsupported J2D type:', pane.type); break;
+            }
+        }
+
+        if (this.data.basePos !== 0) { console.warn('Untested J2D feature'); }
+        if (this.data.rot !== 0) { console.warn('Untested J2D feature'); }
+    }
+
+    // NOTE: Overwritten by child classes 
+    public drawSelf(renderInstManager: GfxRenderInstManager, viewerRenderInput: ViewerRenderInput, ctx2D: J2DGrafContext, offsetX: number, offsetY: number) { }
```
magcius, line 259: Should this be abstract?

src/Common/JSYSTEM/J2D.ts
```diff
@@ -0,0 +1,460 @@
...
+    Left = (1 << 3),
+};
+
+
+// TODO: Move and reorganize
+export class J2DGrafContext {
+    private sceneParams = new SceneParams();
+
+    constructor(device: GfxDevice) {
+        // @NOTE: The y axis is inverted by this ortho matrix. Gamecube origin is top-left, ours is bottom left.
```
magcius, line 60: This seems wrong? The clip space should be the same between both GameCube and noclip, with 0, 0 being the bottom left. The viewport space is different, but that's not involved here.

src/Common/JSYSTEM/J2D.ts
```diff
@@ -0,0 +1,460 @@
...
+            v0 = 0.5 - (this.drawDimensions[1] / texDimensions[1]) / 2.0;
+            v1 = 0.5 + (this.drawDimensions[1] / texDimensions[1]) / 2.0;
+        }
+
+
+        this.sdraw.setVtxDesc(GX.Attr.POS, true);
+        this.sdraw.setVtxDesc(GX.Attr.TEX0, true);
+        this.sdraw.setVtxDesc(GX.Attr.CLR0, true);
+
+        // @NOTE: Y positions are inverted, to account for the projection matrix Y-flip. Gamecube origin is top-left, ours is bottom left.
```
magcius, line 407: Again, this seems sketchy.

src/ZeldaWindWaker/d_place_name.ts
```diff
@@ -0,0 +1,167 @@
...
+    GanonsTower,
+}
+
+export const enum PlacenameState {
+    Init,
+    Hidden,
+    Visible,
+}
+
+export function updatePlaceName(globals: dGlobals) {
```
magcius, line 41: Could you namespace this, like, `dPn__update()`?

src/Common/JSYSTEM/J2D.ts
```diff
@@ -0,0 +1,460 @@
...
+}
+//#endregion
+
+//#region J2DPicture
+export class J2DPicture extends J2DPane {
+    public override data: PIC1;
+
+    private sdraw = new TSDraw(); // TODO: Time to move TSDraw out of Mario Galaxy?
+    private materialHelper: GXMaterialHelperGfx;
+    private tex: BTIData;
```
magcius, line 323:
```suggestion
    private tex: BTIData | null = null;
```

src/Common/JSYSTEM/J2D.ts
themikelester: Oo that's good to know. I'm down with that plan, I'll rename...

src/Common/JSYSTEM/J2D.ts
themikelester: I don't think so, because the J2DPane can't be abstract. I've added to that comment to make it a bit more clear though.

src/Common/JSYSTEM/J2D.ts
themikelester: Hmm, I'll look more into this...

src/ZeldaWindWaker/d_place_name.ts
themikelester: Oh I'd seen that pattern around and wondered about it, now it makes sense. Yes I'll follow that pattern.

src/Common/JSYSTEM/J2D.ts
themikelester: OK after double checking, it _does_ seem like J2D has its origin in the top left. Here's the relevant data from the J2DPicture which controls the positioning for the Wind Waker place name:
```
{
  basePos: 0
  h: 40
  w: 360
  x: 38
  y: 66
}
```

And a screenshot from the original game:
Screenshot 2024-12-15 at 6 04 44 PM

A basePos of 0 doesn't manipulate the origin or the coordinate system at all. I also check that the one parent of the pane doesn't do anything to manipulate the coordinate system. 

I could always adjust this later once we have more use cases for J2D. It should become obvious very fast if this coordinate system is wrong.

magcius: Right, so this is working because when J2D creates the orthographic matrix, it calls C_MTXOrtho like this:

https://github.com/zeldaret/tww/blob/ced4f69ee7af0d6d2e5c60d925bce3854954d86a/src/JSystem/J2DGraph/J2DOrthoGraph.cpp#L27

The arguments to C_MTXOrtho are top, bottom, left, right. So this is putting the top at the origin Y, and the bottom at the size's Y. It has nothing to do with the GameCube's clip space, it's just how J2DOrthoGraph chooses to set up its coordinate system, which the projection matrix will then map to the typical clip space unit cube, which is the same between GameCube and noclip*.

Note that noclip's projectionMatrixForCuboid takes the arguments in a different order, it uses left, right, bottom, top (which matches the glOrtho convention).

* Note that GameCube and noclip differ in clip-space along Z (GameCube uses the unconventional -1...0, and noclip either uses 0...1 or -1..1 depending on backend and availability of the EXT_clip_control extension). But that doesn't concern you here; for X and Y they're identical.

src/Common/JSYSTEM/J2Dv1.ts
```diff
@@ -0,0 +1,475 @@
...
+        mb.setUsePnMtxIdx(false);
+        this.materialHelper = new GXMaterialHelperGfx(mb.finish());
+    }
+}
+//#endregion
+
+//#region J2DScreen
+export class J2DScreen extends J2DPane {
+    public color: Color
+    private static defaultCtx: J2DGrafContext;
```
magcius, line 456: Please do not use module-scope or static variables in noclip code if they hold onto other objects. They will outlive the scene, and will be kept in memory even if the user switches to another game. Pass things through, always.

magcius (changes requested):

src/Common/JSYSTEM/J2Dv1.ts
themikelester: Oo right, and you've told me that before. My bad! And a good reminder for the reasoning. I'll fix this up...

magcius (approved):

PR #736, opened 2024-12-20
Wind Waker: Add black bars to top and bottom during demos

themikelester:
zww_stolensister_2024-12-20T15_32_47 407Z

These bars disappear when the demo is paused or when viewing a normal scene. They animate in and out using the same logic as the game. 

More interestingly, I had to add a few camera-related member variables, which made it seem like time to create `dCamera_c`. For simplicity I left it in `Main.ts`. I also implemented the scissoring as simply as I could, which is a bit hacky. We _could_ introduce `dDlst_window_c` which would get set on each draw list. Then `dCamera_c` would just modify the window during its update. Thoughts?

### DIFF
--- a/src/ZeldaWindWaker/Main.ts
+++ b/src/ZeldaWindWaker/Main.ts
@@ -267,7 +267,7 @@ class dCamera_c {
         
         // From dCamera_c::CalcTrimSize()
         // Animate up to the trim size for the current mode.
-        if(this.cameraMode == CameraMode.Cinematic) {
+        if(this.cameraMode === CameraMode.Cinematic) {
             this.trimHeight += (dCamera_c.trimHeightCinematic - this.trimHeight) * 0.25;
         } else {
             this.trimHeight += -this.trimHeight * 0.25;
--- a/src/ZeldaWindWaker/Main.ts
+++ b/src/ZeldaWindWaker/Main.ts
@@ -202,6 +202,69 @@ export class dGlobals {
     }
 }
 
+class dCamera_c {
+    // For people to play around with.
+    public cameraFrozen = false;
+
+    public execute(globals: dGlobals) {
+        
+    }
+
+    public draw(globals: dGlobals, viewerInput: Viewer.ViewerRenderInput) {
+        // Near/far planes are decided by the stage data.
+        const stag = globals.dStage_dt.stag;
+
+        // Pull in the near plane to decrease Z-fighting, some stages set it far too close...
+        let nearPlane = Math.max(stag.nearPlane, 5);
+        let farPlane = stag.farPlane;
+
+        // noclip modification: if this is the sea map, push our far plane out a bit.
+        if (globals.stageName === 'sea')
+            farPlane *= 2;
+
+        globals.camera.setClipPlanes(nearPlane, farPlane);
+
+        // noclip modification: if we're paused, allow noclip camera control during demos
+        const isPaused = viewerInput.deltaTime === 0;
+
+        // TODO: Determine the correct place for this
+        // dCamera_c::Store() sets the camera params if the demo camera is active
+        const demoCam = globals.scnPlay.demo.getSystem().getCamera();
+        if (demoCam && !isPaused) {
+            let viewPos = globals.cameraPosition;
+            let targetPos = vec3.add(scratchVec3a, globals.cameraPosition, globals.cameraFwd);
+            let upVec = vec3.set(scratchVec3b, 0, 1, 0);
+            let roll = 0.0;
+
+            if (demoCam.flags & EDemoCamFlags.HasTargetPos) { targetPos = demoCam.targetPosition; }
+            if (demoCam.flags & EDemoCamFlags.HasEyePos) { viewPos = demoCam.viewPosition; }
+            if (demoCam.flags & EDemoCamFlags.HasUpVec) { upVec = demoCam.upVector; }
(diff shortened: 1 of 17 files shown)


magcius:
I'm not sure I want this. The original letterboxing effect was there to frame things in a more 16:9 ratio. We already allow users to resize the window to see the same cutscene in different aspect ratios, meaning that we're already not going to get the original viewing proportions correct. So I don't see a major reason to add the letterboxing effect, and I think it goes against the goals of noclip as being a way to "step outside" the game.

themikelester:
Aww, I was afraid you'd say that, but I really like it. When you pause (which allows you to move the camera), they animate it away and it makes it clear that you're stepping outside the game. But when it's playing it does make it seem more cinematic. 

But of course I really don't want to mess with your vision for noclip. If this doesn't fit with that vision, then I'll leave it out.

magcius:
I could be convinced otherwise, but I'd need to spend some time with it.

src/ZeldaWindWaker/Main.ts
```diff
@@ -202,6 +202,94 @@ export class dGlobals {
...
+
+    private trimHeight = 0;
+    private cameraMode: CameraMode = CameraMode.Default;
+    private scissor = vec4.create();
+
+    private static trimHeightCinematic = 65.0;
+
+    public execute(globals: dGlobals, viewerInput: Viewer.ViewerRenderInput) {
+        // Near/far planes are decided by the stage data.
+        const stag = globals.dStage_dt.stag;
```
magcius, line 222: I like the idea of splitting dCamera_c out to a separate class, but I believe we should replace `globals.camera` with it. Search for setupFromCamera to see other examples of how I do this elsewhere.

src/ZeldaWindWaker/Main.ts
```diff
@@ -202,6 +202,94 @@ export class dGlobals {
...
+            this.cameraMode = CameraMode.Cinematic;
+            globals.context.inputManager.isMouseEnabled = false;
+        } else {
+            this.cameraMode = CameraMode.Default;
+            globals.context.inputManager.isMouseEnabled = true;
+        }
+        
+        // From dCamera_c::CalcTrimSize()
+        // Animate up to the trim size for the current mode.
+        if(this.cameraMode == CameraMode.Cinematic) {
```
magcius, line 270: Triple equals

src/ZeldaWindWaker/Main.ts
```diff
@@ -202,6 +202,94 @@ export class dGlobals {
...
+            this.trimHeight += (dCamera_c.trimHeightCinematic - this.trimHeight) * 0.25;
+        } else {
+            this.trimHeight += -this.trimHeight * 0.25;
```
magcius, line 273: Should these respect delta time? It's a bit weird to see the black bars animate if you pause, but also that's what should happen.

src/ZeldaWindWaker/d_place_name.ts
```diff
@@ -114,7 +116,10 @@ export class d_place_name extends msg_class {
 
     public override draw(globals: dGlobals, renderInstManager: GfxRenderInstManager, viewerInput: ViewerRenderInput): void {
         renderInstManager.setCurrentList(globals.dlst.ui[0]);
-        this.screen.draw(renderInstManager, viewerInput, null);
+
+        // @TODO: This will be needed when drawing all Wind Waker 2D elements. It should be in dComIfG_play_c. 
+        const ctx = new J2DGrafContext(globals.context.device, 0.0, 0.0, 608.0, 448.0, -1.0, 0.0);
+        this.screen.draw(renderInstManager, viewerInput, ctx);
```
magcius, line 122: Mind removing the `| null` and fallback in J2DScreen as well?

magcius (commented):
Please also make sure to replace the clear color from TransparentBlack to OpaqueBlack in Main.ts to ensure that the black bars work correctly, since you're now relying on that.

src/ZeldaWindWaker/Main.ts
themikelester: Ah that's something that I missed. Let me see if I can get them back to the correct speed while keeping the pause animation...

src/ZeldaWindWaker/Main.ts
themikelester: Oops my bad!

src/ZeldaWindWaker/d_place_name.ts
themikelester: The 640x480 ortho fallback is in the original J2D, Wind Waker just overrides it with this 608 x 448 one (Do you know why they chose those numbers? Also why is the EFB 528 high?). But yea I'll remove it since I suspect most games are going to be overriding.

themikelester:
What do you think about adding a "Demo Options" UI element? We could add letterboxing as a toggle and stick some play speed / scrubbing controls in there too (after looking through Studio, I think it'd be tough to integrate the scrubbing UI)

src/ZeldaWindWaker/Main.ts
themikelester: Done! It turned out to be a bit messier than the other users of setFromCamera because there are several places in src/Common/ and src/gx/ that use `viewerInput.camera` directly. To work around this without doing some bigger refactoring, I made `dCamera_c` extend `Camera`. 

I think we'd be better off if Camera wasn't used anywhere in src/Common, and we just passed in matrices, frustums, etc or abstracted whatever is using those up to a higher level. But take a look through this and let me know what you think . And feel free to either make changes directly in the branch too.

PR #737, opened 2024-12-24
Migrate to rsbuild

encounter:
[rsbuild](https://rsbuild.dev/) wraps rspack and provides a more batteries-included solution. This provides a large reduction in Node deps and a simpler configuration schema.

I added various comments to `rsbuild.config.ts`, let me know if anything is unclear.

~~Note: the `data` dir was moved to `public/data`, along with `dist/.htaccess` -> `public/.htaccess`. This works better with the built-in dev server and production build logic. If undesired, `public/data` could be a symlink to `../data`, so that devs don't have to move their dir.~~
`public/data/` is now symlinked to `data/`.


src/DataFetcher.ts
```diff
@@ -116,7 +116,7 @@ class DataFetcherRequest {
     public async start() {
         this.started = true;
 
-        if (this.cache !== null) {
+        if (!IS_DEVELOPMENT && this.cache !== null) {
```
encounter, line 119: I added this because it will aggressively cache in dev server mode, even when using empty cache + reload

PR #739, opened 2025-01-01
Wind Waker: Particle system improvements 

themikelester:
- Added Simple particle system (shares emitters for common particle types, eg. flame) 
  - Removed a few TODOs and HACKs related to this (more to come)
  - Allow particles for which `(userID & 0x4000)` is true to be drawn in a particle group other than `Projection`/`Indirect`
- Introduced ParticleGroup enum. Partially determines partical draw order, and in the future some other state such as the current view matrix
- Add support for 2D/UI particle groups
  - These emitters expect different view/proj matrices (ortho, 640x480), and to be drawn later than the 3D emitters
- Fixes incorrect Y position of flame emitters (d_a_ep). Simple particles ignore the emitter translation from their definition, resulting in flame emitters being 20 units lower which matches up with the original game (see flame positioning against torches)

zww_title_2025-01-03T19_05_20 592Z

 
zww_Cave01_2025-01-03T17_23_11 201Z

Impetus here is that the emitters used in d_a_title need to draw into a new ParticleGroup that supports ortho view/proj matrices. Since those emitters have their 0x4000 ID bit set, fixing up the `simple` system seemed appropriate.

"Bomb Island" and "Beedle's Shop" are good scene's to analyze `d_a_ep` (flames) which are the first actor to use the new simple system. "Dragon Roost Cavern" features `d_a_obj_flame` actors that have `useSimpleEm` set. The "Title Screen" demo can be used to test 2D emitters.

### DIFF
--- a/src/Common/JSYSTEM/J2Dv1.ts
+++ b/src/Common/JSYSTEM/J2Dv1.ts
@@ -81,8 +81,6 @@ const enum J2DUVBinding {
 
 export class J2DGrafContext {
     private clipSpaceNearZ: GfxClipSpaceNearZ;
-    private frustum = new Frustum();
-
     public sceneParams = new SceneParams();
     public viewport = vec4.create();
     public ortho = vec4.create();
@@ -121,17 +119,6 @@ export class J2DGrafContext {
         projectionMatrixForCuboid(this.sceneParams.u_Projection, left, right, bottom, top, this.near, this.far);
         projectionMatrixConvertClipSpaceNearZ(this.sceneParams.u_Projection, this.clipSpaceNearZ, GfxClipSpaceNearZ.NegativeOne);
 
-    }
-
-    /**
-     * noclip modification:
-     * Given a view matrix, return a Frustum which can be used for culling. 
-     * In the original game, frustums are generated from proj matrices alone. But noclip expects viewProj frustums. 
-     */
-    public getFrustumForView(viewFromWorldMatrix: mat4): Frustum {
-        const clipFromWorldMatrix = mat4.mul(scratchMat, this.sceneParams.u_Projection, viewFromWorldMatrix);
-        this.frustum.updateClipFrustum(clipFromWorldMatrix, this.clipSpaceNearZ);
-        return this.frustum;
     }
 
     public setOnRenderInst(renderInst: GfxRenderInst): void {
--- a/src/Common/JSYSTEM/J2Dv1.ts
+++ b/src/Common/JSYSTEM/J2Dv1.ts
@@ -17,6 +17,7 @@ import { BTIData } from "./JUTTexture.js";
 import { GXMaterialBuilder } from "../../gx/GXMaterialBuilder.js";
 import { mat4, vec2, vec4 } from "gl-matrix";
 import { GfxRenderCache } from "../../gfx/render/GfxRenderCache.js";
+import { Frustum } from "../../Geometry.js";
 
 const materialParams = new MaterialParams();
 const drawParams = new DrawParams();
@@ -80,10 +81,11 @@ const enum J2DUVBinding {
 
 export class J2DGrafContext {
     private clipSpaceNearZ: GfxClipSpaceNearZ;
+    private frustum = new Frustum();
+
     public sceneParams = new SceneParams();
     public viewport = vec4.create();
     public ortho = vec4.create();
-
     public near: number;
     public far: number;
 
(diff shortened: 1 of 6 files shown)


src/Common/JSYSTEM/J2Dv1.ts
```diff
@@ -118,6 +120,18 @@ export class J2DGrafContext {
...
+    }
+
+    /**
+     * noclip modification:
+     * Given a view matrix, return a Frustum which can be used for culling. 
+     * In the original game, frustums are generated from proj matrices alone. But noclip expects viewProj frustums. 
+     */
+    public getFrustumForView(viewFromWorldMatrix: mat4): Frustum {
+        const clipFromWorldMatrix = mat4.mul(scratchMat, this.sceneParams.u_Projection, viewFromWorldMatrix);
+        this.frustum.updateClipFrustum(clipFromWorldMatrix, this.clipSpaceNearZ);
```
magcius, line 133: It seems rather unlikely to me that you need a Frustum for 2D / UI rendering? The view matrix is also frequently identity in 2D rendering. What is this used for?

src/ZeldaWindWaker/Main.ts
```diff
@@ -403,18 +403,37 @@ export class WindWakerRenderer implements Viewer.SceneGfx {
...
                 globals.particleCtrl.setDrawInfo(globals.camera.viewFromWorldMatrix, globals.camera.clipFromViewMatrix, texPrjMtx, globals.camera.frustum);
-                renderInstManager.setCurrentList(dlst.effect[group]);
+                renderInstManager.setCurrentList(dlst.effect[group == ParticleGroup.Projection ? EffectDrawGroup.Indirect : EffectDrawGroup.Main]);
                 globals.particleCtrl.draw(device, this.renderHelper.renderInstManager, group);
             }
+
+            // From mDoGph_Painter(). Draw the 2D particle groups with different view/proj matrices. 
+            {
+                const orthoCtx = this.globals.scnPlay.currentGrafPort;
+                const viewMtx = mat4.fromTranslation(scratchMatrix, [orthoCtx.aspectRatioCorrection * 320, 240, 0]);
```
magcius, line 422:
```suggestion
                setMatrixTranslation(scratchMatrix, orthoCtx.aspectRatioCorrect * 320, 240, 0);
```

src/Common/JSYSTEM/J2Dv1.ts
themikelester: The current JPA implementation uses `JPADrawInfo.frustum` for per-particle culling. In the original JPA, that frustum is constructed from the proj matrix. In ours it's constructed from the viewProj. I could have done this directly in Main.ts during `JPADrawInfo` setting, but seemed cleaner to add this convenience method.

src/ZeldaWindWaker/Main.ts
themikelester: `setMatrixTranslation` only modifies the translation, if scratchMatrix had a garbage rotation it'd carry over. Setting the whole matrix seems safer. Should I be avoiding  `mat4.fromTranslation()`? I could use `MtxTrans` instead, it's just more cumbersome.

src/Common/JSYSTEM/J2Dv1.ts
magcius: Original JPA doesn't do any frustum culling. In the simple callback, mDoLib_clipper_c is used to do frustum culling [here](https://github.com/zeldaret/tww/blob/4a0a89501d765c5682e76ed5e05cd45648977614/src/d/d_particle.cpp#L454), and it passes the view-matrix along. We rotate our planes rather than the AABB corners, but the same info is all there.

I can't imagine *any* context where J2D needs a Frustum or a view matrix. JPADrawInfo supports passing a null frustum, I would do that for 2D rendering. If you really want a Frustum (you're rendering lots of 2D particles off-screen for some reason), then I would rely on the view matrix being identity, as it should be during 2D rendering, and update the frustum inside setPort.

src/ZeldaWindWaker/Main.ts
magcius: computeModelMatrixT, then. It's not a big deal, but it avoids the garbage generated by the list.

src/Common/JSYSTEM/J2Dv1.ts
themikelester: Ahhh that's a much better idea! Fixed.

PR #744, opened 2025-01-19
Unity fixes + decrunching

wgreenberg:
sorry for the weird branch name, i'm trying out `jj` which doesn't _do_ branches.


rust/Cargo.toml
```diff
@@ -41,3 +41,4 @@ nalgebra-glm = "0.19.0"
 rand = "0.8.5"
 getrandom = { version = "0.2.15", features = ["js"] }
 noclip-macros = { version = "*", path = "./noclip-macros" }
+texture2ddecoder = { git = "https://github.com/wgreenberg/texture2ddecoder" }
```
magcius, line 44: Interesting, I might want to replace https://github.com/magcius/noclip.website/blob/main/src/Common/bc_texture.ts with this eventually, possibly.

PR #747, opened 2025-01-22
new deku

magcius:


rust/Cargo.toml
```diff
@@ -25,7 +25,7 @@ opt-level = "s"
 [dependencies]
 byteorder = "1.4.3"
 console_error_panic_hook = "0.1.7"
-deku = { version = "0.16.0", features = ["logging"] }
+deku = { version = "0.18.0", features = ["logging"] }
```
wgreenberg, line 28: looks like there's `18.1` now w/ some minor bugfixes

rust/src/unity/types/binary.rs
```diff
@@ -220,21 +221,12 @@ pub struct ByteArray {
...
-        Ok((
-            rest,
-            Self {
-                data: bytes.to_vec(),
-            },
-        ))
+impl<'a, Ctx> DekuReader<'a, Ctx> for ByteArray where Ctx: Copy {
+    fn from_reader_with_ctx<R: deku::no_std_io::Read + deku::no_std_io::Seek>(reader: &mut Reader<R>, _ctx: Ctx) -> Result<Self, DekuError> {
+        let count = i32::from_reader_with_ctx(reader, ())? as usize;
+        let mut buf = vec![0x00; count];
```
wgreenberg, line 227: can you just do `let mut buf = Vec::with_capcity(count);` to avoid zeroing the buffer?

rust/src/unity/types/common.rs
```diff
@@ -21,28 +23,22 @@ pub struct UnityArray<T> {
...
-            Ok((rest, UnityArray {
-                values,
-            }))
+impl<'a, T, Ctx> DekuReader<'a, Ctx> for UnityArray<T> where T: DekuReader<'a, Ctx>, Ctx: Clone {
+    fn from_reader_with_ctx<R: deku::no_std_io::Read + deku::no_std_io::Seek>(reader: &mut Reader<R>, ctx: Ctx) -> Result<Self, DekuError> {
+        let count = i32::from_reader_with_ctx(reader, ())? as usize;
+        let mut values = Vec::new();
+        for _ in 0..count {
+            let value = T::from_reader_with_ctx(reader, ctx.clone())?;
+            values.push(value);
```
wgreenberg, line 37:
```suggestion
            values.push(T::from_reader_with_ctx(reader, ctx.clone())?);
```

rust/src/unity/types/common.rs
```diff
@@ -58,30 +54,23 @@ pub struct Map<K, V> {
...
-            }))
+    fn from_reader_with_ctx<R: deku::no_std_io::Read + deku::no_std_io::Seek>(reader: &mut Reader<R>, ctx: Ctx) -> Result<Self, DekuError> {
+        let count = i32::from_reader_with_ctx(reader, ())?;
+        let mut keys = Vec::new();
+        let mut values = Vec::new();
+        for _ in 0..count {
+            let key = K::from_reader_with_ctx(reader, ctx.clone())?;
+            let value = V::from_reader_with_ctx(reader, ctx.clone())?;
+            keys.push(key);
+            values.push(value);
```
wgreenberg, line 68:
```suggestion
            keys.push(K::from_reader_with_ctx(reader, ctx.clone())?);
            values.push(V::from_reader_with_ctx(reader, ctx.clone())?);
```

rust/src/unity/types/common.rs
```diff
@@ -235,21 +224,19 @@ impl From<Packedi32Vec> for Vec<i32> {
...
-            }))
+impl<'a, Ctx> DekuReader<'a, Ctx> for Packedi32Vec where Ctx: Clone {
+    fn from_reader_with_ctx<R: deku::no_std_io::Read + deku::no_std_io::Seek>(reader: &mut Reader<R>, _ctx: Ctx) -> Result<Self, DekuError> {
+        let num_items = u32::from_reader_with_ctx(reader, ())? as usize;
+        let byte_array_count = u32::from_reader_with_ctx(reader, ())? as usize;
+        let mut buf = vec![0x00; byte_array_count];
+        reader.read_bytes(byte_array_count, &mut buf)?;
+        let bit_size = u8::from_reader_with_ctx(reader, ())?;
+        reader.skip_bits(3*8)?; // align
+        let data = unpack_i32s(&mut buf, num_items as usize, bit_size as usize)?;
```
wgreenberg, line 235: since `unpack_i32s` is creating a cursor/reader anyway, can we just seek this reader back to where the byte array starts, then pass it through here?

rust/src/unity/types/wasm.rs
```diff
@@ -14,9 +16,10 @@ macro_rules! define_create {
         #[wasm_bindgen(js_class = $u)]
         impl $t {
             pub fn create(version: UnityVersion, data: &[u8]) -> Result<$t, String> {
-                let bitslice = BitSlice::from_slice(data);
-                match binary::$t::read(&bitslice, version) {
-                    Ok((_, value)) => Ok(value.into()),
+                let mut cursor = Cursor::new(&data);
```
wgreenberg, line 19: unnecessary reference
```suggestion
                let mut cursor = Cursor::new(data);
```

rust/src/wow/db.rs
```diff
@@ -128,23 +127,34 @@ impl Wdc4Db2File {
...
-    for bit_num in bit_offset..bit_offset + num_bits {
-        let byte_index = bit_num >> 3;
-        let bit_index = 7 - (bit_num % 8);
-        if bits[byte_index * 8 + bit_index] {
-            result |= 1 << (bit_num - bit_offset);
+fn from_u32<T>(v: u32) -> Result<T, DekuError>
+    where for<'a> T: DekuReader<'a, ()>
+{
+    let mut v_bytes = v.to_le_bytes();
+    let mut cursor = Cursor::new(&mut v_bytes);
```
wgreenberg, line 134:
```suggestion
    let mut cursor = Cursor::new(v_bytes);
```

rust/src/wow/db.rs
```diff
@@ -128,23 +127,34 @@ impl Wdc4Db2File {
...
     }
-    result
-}
 
-fn from_u32<T>(v: u32) -> Result<T, DekuError>
-    where for<'a> T: DekuRead<'a, ()>
-{
-    let v_bytes = v.to_le_bytes();
-    let (_, result) = T::read(BitSlice::from_slice(&v_bytes), ())?;
+    reader.seek(std::io::SeekFrom::Start(old)).unwrap();
```
wgreenberg, line 157: not entirely sure, but can you use `reader.seek_last_read()` here?

rust/src/wow/db.rs
```diff
@@ -128,23 +127,34 @@ impl Wdc4Db2File {
...
+{
+    let mut v_bytes = v.to_le_bytes();
+    let mut cursor = Cursor::new(&mut v_bytes);
+    let mut reader = Reader::new(&mut cursor);
+    T::from_reader_with_ctx(&mut reader, ())
+}
+
+fn read_field_to_u32<R: deku::no_std_io::Read + deku::no_std_io::Seek>(reader: &mut Reader<R>, field_offset_bits: usize, field_size_bits: usize) -> Result<u32, DekuError> {
+    // Assumes the reader points to the start of the record
+    let old = reader.seek(std::io::SeekFrom::Current(0i64)).unwrap();
```
wgreenberg, line 141: [looks like they implemented `From<Error> for DekuError`](https://docs.rs/deku/latest/deku/error/enum.DekuError.html#impl-From%3CDekuError%3E-for-Error), so i think you should be able to just do:

```suggestion
    let old = reader.seek(std::io::SeekFrom::Current(0))?;
```

rust/src/wow/db.rs
```diff
@@ -463,14 +475,18 @@ pub struct LightResult {
...
     pub coords: Vec3,
+    #[deku(reader = "db2.read_field(deku::reader, 1)")]
     pub falloff_start: f32,
+    #[deku(reader = "db2.read_field(deku::reader, 2)")]
     pub falloff_end: f32,
+    #[deku(reader = "db2.read_field(deku::reader, 3)")]
     pub map_id: u16,
+    #[deku(reader = "db2.read_field(deku::reader, 4)")]
     light_param_ids: [u16; 8],
-    pub unk: u16,
```
wgreenberg, line 473: interesting, does this not throw things off?

wgreenberg (changes requested):

rust/src/wow/db.rs
magcius: Do I need to import a trait or something?

```
error[E0277]: `?` couldn't convert the error to `deku::DekuError`
   --> src\wow\db.rs:215:61
    |
215 |         let old = reader.seek(std::io::SeekFrom::Current(0))?;
    |                                                             ^ the trait `From<std::io::Error>` is not implemented for `deku::DekuError`, which is required by `std::result::Result<String, deku::DekuError>: FromResidual<std::result::Result<Infallible, std::io::Error>>`
    |
    = note: the question mark operation (`?`) implicitly performs a conversion on the error value using the `From` trait
    = help: the following other types implement trait `From<T>`:
              <deku::DekuError as From<Infallible>>
              <deku::DekuError as From<TryFromSliceError>>
              <deku::DekuError as From<std::num::TryFromIntError>>
    = note: required for `std::result::Result<String, deku::DekuError>` to implement `FromResidual<std::result::Result<Infallible, std::io::Error>>` 
```

rust/src/wow/db.rs
magcius: we're always advancing according to header.record_size and seeking within the record ourselves now

rust/src/wow/db.rs
magcius: don't think so, we're seeking and then skipping

rust/src/wow/db.rs
wgreenberg: oh i'm dumb, they implemented `From<DekuError> for Error` lol. looks like one thing you could do is

```
    let old = reader.seek(std::io::SeekFrom::Current(0))
        .map_err(|err| DekuError::Io(error.kind()))?;
```

magcius:
I think this should be ready for review. I fixed both WoW and Unity loading.

PR #757, opened 2025-04-09
The SpongeBob SquarePants Movie

seilweiss:
This adds partial support for The SpongeBob SquarePants Movie for Xbox, which runs on the same engine as Battle for Bikini Bottom. All levels from the game are viewable, however some objects, most notably the player and NPC/enemies, are not yet visible.

Some improvements to the game engine:
- Teleport boxes are now visible
- BfBB's beta HIP/HOP format is supported - Patrick's unused dream level can now be viewed

Improvements to the RenderWare reimplementation:
- An attempt to separate Noclip-specific code into its own "platform layer" (rwgfx.ts) that handles all the low-level rendering and exposes a simple OpenGL 1.1-style API that makes it easier to recreate RenderWare's pipelines
- Support for mipmapped and DXT-compressed Xbox textures
- Fixed(?) frustum culling


magcius:
The biggest note I have is that the rewrite of rwgfx where everything uses global state is a bit ugly. I intentionally separate out state like that into discrete parts (e.g. GXMaterialBuilder) for a reason.

PR #759, opened 2025-04-14
Add Plus! for XP

Rezmason:
The "[Plus! for XP](https://en.wikipedia.org/wiki/Microsoft_Plus!#Microsoft_Plus!_SuperPack_for_Windows_XP)" Microsoft software package in 2001 included several screensavers, three of which shared an asset system and rendering engine:
- Mercury Pool
image
- Robot Circus
image
- Sand Pendulum
image

This PR adds support for these screensavers, including:
- TIFF support
- baked vertex lighting
- spherical environment texture mapping
- full parser of a proprietary scene descriptor format, "SCX"
- bespoke scene graph
- bespoke animation system
- cameras animated on paths
- simulated water
- simulated sand art
- minor automatic mesh repair


magcius (commented):
Sorry it's taken me so long to get around to reviewing this, I've been busy with a lot of things. There's a lot of things I want to mention, but first off, I'm not happy with adding jpeg-js as a dependency; tiff support I think I can accept, though I might prefer a leaner implementation.

To replace jpeg-js, use the browser's native support for  loading and uploading; look at loadPNGData in the Diddy Kong Racing code, for instance. (Once it's more widely supported, it would be nice to switch to either [ImageDecoder](https://developer.mozilla.org/en-US/docs/Web/API/ImageDecoder) or [ImageBitmap](https://developer.mozilla.org/en-US/docs/Web/API/ImageBitmap)).

In general, there's use of two-space indents here; while not a hard and fast rule, I would prefer you to match the rough style of the rest of the codebase, it makes it a lot easier for me to read through.

The other general complaint is I see you doing a lot of work at module-time -- whether that's parsing large strings into arrays, or calculating large arrays. This work will be done at page initialization time and blocks the time to first render; there are globals that you would like initialized, I would prefer if you had a "Globals" class that had a lot of this data that you instantiated and passed around in your renderer.

Additionally, I notice a strong functional style here; lots of map/reduce, and lots of records and hashmaps being passed around. While I don't demand that you adopt a more traditional OOP approach, it's something that I suspect might be a bit cleaner in terms of resource ownership than the current approach.

src/PlusForXP/scx/parser.ts
```diff
@@ -0,0 +1,246 @@
...
+	"keys": "keyframes",
+	"anim": "animations"
+};
+
+export const parse = async (scxBytes: Uint8Array) : Promise<SCX.Scene> => {
+	const writer = new Writer();
+	const len = scxBytes.length;
+	for (let i = 0; i < len; i++) {
+		const byte = scxBytes[i];
+		const token = tokenTable[byte];
```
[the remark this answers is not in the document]
Rezmason: These screensavers supported the binary `SCX` and human-readable `SCN` formats; the parser began as a script to convert `SCX` to `SCN`. That's why there are also several tokens that this project never uses— the parser supports _any_ `SCX` file.

If you like, I can take another crack at it, but I don't think its complexity will reduce much. I'd probably still use lookup tables instead of switch statements. A rewrite might end up being slightly more type safe, though I'm not sure.

src/PlusForXP/program.ts
```diff
@@ -0,0 +1,87 @@
...
+  v_DiffuseColor = min(a_DiffuseColor, 1.0);
+  v_DiffuseTexCoord = a_TexCoord;
+
+  
+  vec3 e = normalize(worldPosition.xyz - u_ViewInverseMatrix[3].xyz);
+  vec3 n = normalize((u_ModelInverseTransposeMatrix * normal).xyz);
+  
+  vec3 r = reflect(e, n);
+  r = (u_EnvMapMatrix * vec4(r, 1.0)).xyz;
+  v_EnvTexCoord = r.xy / (2.0 * length(r)) + 0.5;
```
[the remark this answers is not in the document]
Rezmason: Close! :) `normalize(r).xy * 0.5 + 0.5`. In it goes!

src/PlusForXP/bake_lights.ts
```diff
@@ -0,0 +1,113 @@
+import { mat3, mat4, vec3 } from "gl-matrix";
+import { SCX } from "./scx/types.js";
+import { Vec3One } from "../MathHelpers.js";
+
+export const bakeLights = (mesh: SCX.Mesh, material:SCX.Shader, worldTransform: mat4, lights: SCX.Light[]) : Float32Array => {
+  lights = [...lights].reverse();
+
+  const useMaterialColors = material.luminance === 0 && (material.blend < 1 || material.texture == null);
```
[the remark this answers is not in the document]
Rezmason: A little context: the original screensavers baked their lighting on the CPU, and this light baking module is the result of months of reverse engineering that original process, with the identical numeric output (within a margin of error comparable to the precision loss from double to float).

I have an early branch where I experimented with a vertex shader implementation, with lights getting passed in as an array of structs and all that. It's possible, but in my opinion, it would make the shader harder to maintain and debug, and the code would stray a bit further from the original.

src/PlusForXP/scx/parser.ts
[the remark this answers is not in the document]
Rezmason: My opinion about the current parser implementation is souring in real time. I'll try again.

src/PlusForXP/scx/parser.ts
```diff
@@ -0,0 +1,246 @@
...
+import { range } from "../../MathHelpers.js";
+import { splitMesh } from "./split_mesh.js";
+
+const littleEndian = (() => {
+	const buffer = new ArrayBuffer(2);
+	new DataView(buffer).setInt16(0, 256, true);
+	return new Int16Array(buffer)[0] === 256;
+})();
+
+const toDataView = (bytes: Uint8Array) => new DataView(
```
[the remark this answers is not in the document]
Rezmason: I'll try to use this in the parser rewrite. Thanks for the heads up!

src/PlusForXP/sandProgram.ts
```diff
@@ -0,0 +1,80 @@
+import { GfxShaderLibrary } from "../gfx/helpers/GfxShaderLibrary";
```
magcius, line 1: Minor nit: the rest of the filenames seem to be named something like foo_bar.ts, but this is fooBar.ts?

src/PlusForXP/sandProgram.ts
```diff
@@ -0,0 +1,80 @@
...
+      vec2 u_LastPendulumPos;
+      vec2 u_PendulumPos;
+      float u_Fade;
+      float u_NumParticles;
+    };
+
+    uniform sampler2D sandTexture;
+
+    #define PI 3.14159265359
+    #define TWO_PI 6.28318530718
```
magcius, line 24: I typically use string interpolation and do `${Math.PI}` and `${MathConstants.TAU}`

src/PlusForXP/simulations/sand_pendulum.ts
```diff
@@ -0,0 +1,437 @@
...
+                this.renderInstListSand.drawOnPassRenderer(renderHelper.renderCache, passRenderer);
+            });
+        });
+
+        const template = renderHelper.pushTemplateRenderInst();
+
+        renderInstManager.setCurrentList(this.renderInstListSand);
+
+        template.setBindingLayouts(Plus4XPSandProgram.bindingLayouts);
+        const gfxProgram = renderInstManager.gfxRenderCache.createProgramSimple(this.sandProgram);
```
magcius, line 409: Mind calling this in the constructor instead of here? Even when cached, `createProgramSimple` isn't the cheapest thing...

src/PlusForXP/simulations/sand_pendulum.ts
```diff
@@ -0,0 +1,437 @@
...
+
+    override setup(device: GfxDevice, world: World): void {
+        this.isGrotto = world.sceneNodesByName.has("Pendulum_SW_Pendulum.scx/Pendulum Arrowhead");
+        this.pendulum = world.sceneNodesByName.get(this.isGrotto ? "Pendulum_SW_Pendulum.scx/Pendulum Arrowhead" : "Pendulum_Pendulum.scx/Pendulum")!;
+        this.sandParticles = world.sceneNodesByName.get("Pendulum_Sand_Particles.scx/_root")!;
+        this.sparkle = world.sceneNodesByName.get("Sparkle.scx/_root")!;
+        this.sparkleSprite = world.sceneNodesByName.get("Sparkle.scx/Plane01")!;
+        this.sparkleSprite.isGhost = true;
+
+        this.sandProgram = preprocessProgramObj_GLSL(device, new Plus4XPSandProgram());
```
magcius, line 132: minor nit but i kinda like having the shader program and user in the same file these days, especially for one-offs, but you do you

src/PlusForXP/renderer.ts
```diff
@@ -0,0 +1,300 @@
...
+
+        renderInstManager.setCurrentList(this.renderInstListMain);
+        this.renderSceneNodeMeshes(renderInstManager, false);
+        this.renderSceneNodeMeshes(renderInstManager, true);
+        renderInstManager.popTemplate();
+        this.renderHelper.prepareToRender();
+    }
+
+    private renderSceneNodeMeshes(renderInstManager: GfxRenderInstManager, ghosts: boolean) {
+        renderInstManager.pushTemplate().setMegaStateFlags({
```
magcius, line 237: in theory you can use the sort key system (or use a separate render list) for ghosts, so you don't need to do this iteration twice

src/PlusForXP/renderer.ts
```diff
@@ -0,0 +1,300 @@
...
+        cameraPanel.customHeaderBackgroundColor = UI.COOL_BLUE_COLOR;
+        cameraPanel.setTitle(UI.EYE_ICON, "Vantage Points");
+        this.cameraSelect = new UI.SingleSelect();
+        this.cameraSelect.setStrings(this.cameras.map((a) => a[0]));
+        this.cameraSelect.onselectionchange = (strIndex: number) => {
+            const choice = this.cameras[strIndex];
+            this.activeCameraName = choice[1];
+            this.lastViewerCameraMatrix = null;
+        };
+        this.cameraSelect.selectItem(1); // TODO: persist through serialize/deserialize
```
magcius, line 117: I would like the camera to be in noclip mode by default.

Rezmason:
> Sorry for the super late review, been busy this past month. A lot of these are nits or suggestions that you don't have to fix if you don't feel like it.

As long as you don't mind, I'm gonna do probably all of these. I'm learning a lot here, and it's part of the fun. 🌟

src/PlusForXP/simulations/sand_pendulum.ts
Rezmason: Embarrassing in hindsight 😅

PR #795, opened 2025-09-22
Added MK64

M-1-RLG:


src/MarioKart64/render.ts
```diff
@@ -0,0 +1,484 @@
...
+        let offs = 0;
+
+        offs += fillVec4(d, offs, renderInput.backbufferHeight / 2);
+        offs += fillColor(d, offs, this.ColorTop);
+        offs += fillColor(d, offs, this.ColorBottom);
+        renderInstManager.submitRenderInst(renderInst);
+    }
+
+    public destroy(device: GfxDevice): void {
+        device.destroyProgram(this.gfxProgram);
```
magcius: Don't destroy things created with the render cache.

src/MarioKart64/render.ts
```diff
@@ -0,0 +1,484 @@
...
+#endif
+}
+`;
+
+    public override frag: string = `
+
+void main() {
+    float y = gl_FragCoord.y;
+
+    if (y >= u_HalfScreenHeight) {
```
magcius: If you used `makeFullscreenVS`, you could use `v_TexCoord` for this instead of having to pass it as a uniform.

src/MarioKart64/render.ts
```diff
@@ -0,0 +1,484 @@
...
+    vec4 u_SkyColorBottom;
+};
+
+#define u_HalfScreenHeight (u_Params.x)
+`;
+
+    public override vert: string = `
+
+void main() {
+    vec2 p;
```
magcius: `GfxShaderLibrary.makeFullscreenVS` ?

src/MarioKart64/render.ts
```diff
@@ -0,0 +1,484 @@
...
+            entry.tile.uls = uls;
+            entry.tile.ult = ult;
+        }
+    }
+
+    public setRenderMode(renderMode: Partial<GfxMegaStateDescriptor>): void {
+        const cullMode = this.megaStateFlags.cullMode;
+
+        this.megaStateFlags = renderMode;
+        this.megaStateFlags.cullMode = cullMode;
```
magcius, line 344: Note that this will set the megaStateFlags for all render objects to the same object, and then overwrite the cullMode.

```suggestion
        const cullMode = this.megaStateFlags.cullMode;

        setMegaStateFlags(this.megaStateFlags, renderMode);
        this.megaStateFlags.cullMode = cullMode;
 ```

File: src/MarioKart64/f3dex.ts [hunk 5]
```diff
@@ -0,0 +1,455 @@
...
+    constructor(public drawCalls: MkDrawCall[], public textureCache: RDP.TextureCache) {
+    }
+}
+
+export class MkRSPState implements F3DEX.RSPStateInterface {
+    private stateChanged: boolean = false;
+    private outputDrawCalls: MkDrawCall[] = [];
+    private currentDrawCall = new MkDrawCall();
+
+    private vertexCache: F3DEX.Vertex[] = [];
```
magcius:
```suggestion
    private vertexCache: F3DEX.Vertex[] = nArray(64, () => new F3DEX.Vertex());
```

src/MarioKart64/courses.ts
```diff
@@ -0,0 +1,8027 @@
...
+            const boneMtx = this.boneMatrices[bone.index];
+
+            let rotX = getTrackFrameValue(anim.rotationTracks[i].x, true);
+            let rotY = getTrackFrameValue(anim.rotationTracks[i].y, true);
+            let rotZ = getTrackFrameValue(anim.rotationTracks[i].z, true);
+
+            computeModelMatrixSRT(boneMtx, 1, 1, 1, rotX, rotY, rotZ, bone.pos[0], bone.pos[1], bone.pos[2]);
+
+            if (i === 0) {
+                this.boneMatrices[1][12] += getTrackFrameValue(anim.translationTrack.x);
```
magcius: Is this not the same as boneMtx?

src/MarioKart64/actors.ts
```diff
@@ -0,0 +1,1043 @@
...
+    public init(renderCache: GfxRenderCache, rspState: MkRSPState): void { }
+
+    public update(deltaTimeFrames: number): void { }
+
+    public prepareToRender(renderInstManager: GfxRenderInstManager, viewerInput: Viewer.ViewerRenderInput): void { }
+
+    public destroy(device: GfxDevice): void { }
+
+    public setShadowMatrix(dst: mat4, scale: number, pos: vec3 = this.pos, rot: vec3 = this.rot): void {
+        this.pos[1] += 2;
```
magcius: I'm guessing you want this to be `pos[1]`, not `this.pos[1]`, in order to catch the case you pass scratchVec3 into it?

src/MarioKart64/utils.ts
```diff
@@ -0,0 +1,281 @@
...
+
+    return (dx * dx + dz * dz) <= (distance * distance);
+}
+
+export function IsTargetInRangeXYZ(targetPos: vec3, cameraPos: vec3, distance: number): boolean {
+    const dx = targetPos[0] - cameraPos[0];
+    const dy = targetPos[1] - cameraPos[1];
+    const dz = targetPos[2] - cameraPos[2];
+
+    return (dx * dx + dy * dy + dz * dz) <= (distance * distance);
```
magcius:
```suggestion
    return vec3.squaredDistance(targetPos, cameraPos) <= distance ** 2;
```

src/MarioKart64/utils.ts
```diff
@@ -0,0 +1,281 @@
...
+        pos[0], pos[1], pos[2]);
+
+    return dst;
+}
+
+const defaultLightDir: vec3 = [0, 0, (120 / 0x7F)];
+export function rotateVectorXY(dst: vec3, rot: vec3): void {
+    const x = defaultLightDir[0];
+    const y = defaultLightDir[1];
+    const z = defaultLightDir[2];
```
magcius: If you wanted to unfold this, then you could easily do so, given that x/y are 0 in the math below.

src/MarioKart64/utils.ts
```diff
@@ -0,0 +1,281 @@
...
+        }
+
+        spawnData.push({ pos: vec3.fromValues(posX, posY, posZ), params, posY: originalPosY });
+    }
+
+    return spawnData;
+}
+
+export function isNearFrameTime(timer: number, targetFrame: number, threshold: number = 0.5): boolean {
+    return Math.abs(timer - targetFrame) < threshold;
```
magcius: Perhaps for something more reliable, you can do something like `crossedTime` which appears in SMG/Source? That should work regardless of stepping "too fast"

src/MarioKart64/utils.ts
```diff
@@ -0,0 +1,281 @@
...
+export function hashFromValues(a: number[]): number {
+    let hash = 0;
+    for (const value of a) {
+        hash = hashCodeNumberUpdate(hash, value);
+    }
+    return hashCodeNumberFinish(hash);
+}
+
+export function mod(val: number, n: number): number {
+    return ((val % n) + n) % n;
```
magcius: Is this functionally different from `mod` that appears in util.ts?

src/MarioKart64/courses.ts
```diff
@@ -0,0 +1,8027 @@
...
+    public deleteObject(): void {
+        this.resetAllStates();
+        this.isInitialized = false;
+    }
+
+
+    public updateTextures(renderCache: GfxRenderCache, rspState: MkRSPState, dramAddr: number, dramPalAddr: number): void {
+        assert(this.modelInst !== undefined);
+
+        const cacheKey = hashFromValues([dramAddr, dramPalAddr]);
```
magcius: You could make the cache key be a string, and then do `${dramAddr}_${dramPalAddr}` ?

src/MarioKart64/courses.ts
```diff
@@ -0,0 +1,8027 @@
...
+}
+
+const bindingLayouts: GfxBindingLayoutDescriptor[] = [
+    { numUniformBuffers: 3, numSamplers: 2, },
+];
+
+export class Mk64Renderer implements Viewer.SceneGfx {
+    public static courseId: CourseId;
+    public static gfxTextureCache: Map<number, GfxTexture> = new Map();
+    public static modelCache = new Map<number, BasicRspRenderer>();
```
magcius: While I understand it can be hard to avoid, I really don't like having static globals, since they can easily leak between games, cause use-after-free issues, leaks, and other issues like that.

src/MarioKart64/courses.ts
```diff
@@ -0,0 +1,8027 @@
...
+        return angleDiff <= tolerance;
+    }
+
+    public IsBehindCamera(pos: vec3): boolean {
+
+        const yawRad = this.cameraYaw;
+        const dx = pos[0] - this.cameraPos[0];
+        const dz = pos[2] - this.cameraPos[2];
+
+        return (dx * Math.sin(yawRad) + dz * Math.cos(yawRad)) < 0;
```
magcius: Why not `dot(this.cameraFwd.xz, vec2(dx, dz)) < 0` ?

src/MarioKart64/courses.ts
```diff
@@ -0,0 +1,8027 @@
...
+    private commonShadow: BasicRspRenderer;
+    private flagMdl: BasicRspRenderer;
+
+    public nearestTrackSection = 0;
+    public nearestPathPointIdx = 0;
+    public driveInReverse: boolean = false;
+    public trackOffsetPosition: vec3 = vec3.create();
+
+
+    // Camera junk... I didn't expect to need all this
```
magcius: I might take an attempt at reducing/removing some of this?

src/MarioKart64/courses.ts
```diff
@@ -0,0 +1,8027 @@
...
+    public static courseId: CourseId;
+    public static gfxTextureCache: Map<number, GfxTexture> = new Map();
+    public static modelCache = new Map<number, BasicRspRenderer>();
+
+    private commonShadow: BasicRspRenderer;
+    private flagMdl: BasicRspRenderer;
+
+    public nearestTrackSection = 0;
+    public nearestPathPointIdx = 0;
+    public driveInReverse: boolean = false;
```
magcius: ???

src/MarioKart64/courses.ts
M-1-RLG: Copied from decomp iirc. A lot of code still needs to be cleaned up for noclip implementation

src/MarioKart64/courses.ts
M-1-RLG: "nearestTrackSection" is the "track section" below the camera. It used to trigger things like the fire breath in bowsers castle, only when you're in section 5.

"nearestPathPointIdx" is the closest path point below the camera. It is used to trigger boos in Banshee boardwalk only in certain areas

"driveInReverse" makes the vehicles in toads turnpike drive in the opposite direction. I don't support mirror mode yet so I forgot about this.

magcius: Yeah, this was meant to be focused on driveInReverse, as it's not written to anywhere.

PR #879, opened 2026-02-12
Bind rust wasm to pseudo package

kaze-xiv:
Added a [paths option in tsconfig.json](https://www.typescriptlang.org/tsconfig/#paths) that imports the rust library as a pseudo package. This removes a bit of the cruft associated with going up and back down the filesystem hierarchy when talking to wasm and removes the burden of knowledge on the magic `pkg` path for library consumers.

I picked the name `noclip-support` to match the crate name but understand that it could be confused for a package on npm named `noclip-support`. There currently isn't one. Some projects use `~` to denote alias imports, so that's an alternative too.


magcius:
What about `noclip-rust-support` or `noclip-support.rs` to clarify? Feel free to rename the crate as well.

PR #889, opened 2026-03-15
Add Top Gear Rally (N64)

Tatsh:
I am getting bored of it over the last 3 days so here is most of the job done. It is placed in the Experimental section because of the following issues.

## Affecting all tracks

- Billboard spline-animated models usually have a 'best' viewing angle, and when not viewed at that angle, the skybox is often in the transparent section of the texture.
- Weather is not properly implemented (commented out). This affects some signage in most tracks.
- Reflection is probably implemented incorrectly.

## Coastline

- Missing lighthouse light and its animation.
- Windmill blades clipped at bottom.

## Jungle

- Missing fire above the heads in the tunnel.
- Some directional signs are just solid colours (yellow ones in this track).
- Road texture is shown incorrect in some places (shortcut and towards end of track).
- Second to last waterfall not animated; part of water body missing.

## Desert

- Light textures on the ceiling do not appear correctly in the second and last long tunnel.
- 'FISH N CHIPS' sign is probably supposed to be mirrored when the backside is shown to camera or it is supposed to be a billboard (as stored in the ROM, it rotates).

## Mountain

- Some signs show a different than expected texture (might be directional signs).
- Most of the checkmark banner textures are mirrored.

## Strip Mine

- Animation splines are traced backwards.
- START banner is cut off on the right.
- Most of the checkmark banner textures are mirrored.

## Season Winner

- Camera flashes are not implemented correctly (original code is not designed for free camera).

### DIFF
--- a/src/Common/N64/RDP.ts
+++ b/src/Common/N64/RDP.ts
@@ -472,13 +472,7 @@ export function translateTileTexture(segmentBuffers: ArrayBufferSlice[], dramAdd
     default:
         throw new Error(`Unknown image format ${tile.fmt} / ${tile.siz}`);
     }
-    // Clone the tile state so the Texture has its own immutable snapshot.
-    // Without this, the Texture holds a mutable reference to DP_TileState
-    // which gets overwritten by subsequent SETTILE commands, corrupting
-    // the sampler wrap modes and texture cache lookups.
-    const snapshotTile = new TileState();
-    snapshotTile.copy(tile);
-    const out = new Texture(snapshotTile, dramAddr, dramPalAddr, tileW, tileH, dst);
+    const out = new Texture(tile, dramAddr, dramPalAddr, tileW, tileH, dst);
 
     return out;
 }
--- a/src/Common/N64/RDP.ts
+++ b/src/Common/N64/RDP.ts
@@ -472,7 +472,13 @@ export function translateTileTexture(segmentBuffers: ArrayBufferSlice[], dramAdd
     default:
         throw new Error(`Unknown image format ${tile.fmt} / ${tile.siz}`);
     }
-    const out = new Texture(tile, dramAddr, dramPalAddr, tileW, tileH, dst);
+    // Clone the tile state so the Texture has its own immutable snapshot.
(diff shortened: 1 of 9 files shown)


src/Common/N64/RDP.ts
```diff
@@ -472,7 +472,13 @@ export function translateTileTexture(segmentBuffers: ArrayBufferSlice[], dramAdd
...
         throw new Error(`Unknown image format ${tile.fmt} / ${tile.siz}`);
     }
-    const out = new Texture(tile, dramAddr, dramPalAddr, tileW, tileH, dst);
+    // Clone the tile state so the Texture has its own immutable snapshot.
+    // Without this, the Texture holds a mutable reference to DP_TileState
+    // which gets overwritten by subsequent SETTILE commands, corrupting
+    // the sampler wrap modes and texture cache lookups.
+    const snapshotTile = new TileState();
+    snapshotTile.copy(tile);
+    const out = new Texture(snapshotTile, dramAddr, dramPalAddr, tileW, tileH, dst);
```
magcius: I don't think that's right? The constructor for Texture copies the tile internally.

src/TopGearRally/render.ts
```diff
@@ -0,0 +1,1654 @@
...
+    envColor: vec4;
+}
+
+/**
+ * Scan raw DL bytes to extract G_SETPRIMCOLOR / G_SETENVCOLOR
+ * state indexed by cumulative triangle output position. Records
+ * the running prim/env colour at each triangle index so that
+ * DrawCallInstances can look up colours by firstIndex.
+ */
+function scanDLColors(
```
magcius: This is fine but you could add gSPSetPrimColor to RSPStateInterface the same way we added it to PokemonSnap. The shared N64 code is a bit of a mess though.

Tatsh: I think this what you meant right? https://github.com/magcius/noclip.website/pull/889/commits/24b683147439a3b25c6dac061970836a1b3c61c1

src/Common/N64/RDP.ts
Tatsh: I removed this change. Was not necessary.

PR #895, opened 2026-04-08
Add Paper Mario: The Origami King, use tegra_swizzle crate

egartley:
Added levels from _Paper Mario: The Origami King_ for the Switch. Like the other Switch games, it's in experimental.

Data is simple to set up, just dump the cart and extract romfs to `data/PaperMarioTOK`.

There's a big list of todo items in the comments of the scenes file. These are just minor fixes, optimizations, and some manual tweaking I'll do later - the game itself is more or less feature complete.

Credit goes to Odyssey and MK8D for the basis of rendering and parsing, and TMSFE for skeletal animations. [Origami Wand](https://github.com/Darxoon/OrigamiWand) helped with parsing some files. Switch Toolbox was an invaluable resource too.

A few technical notes:

- Load times can be a little rough, especially for the larger levels like Toad Town. This is due to needing to decompress the ASTC textures, which the game uses heavily
  - The [WebGL extension](https://developer.mozilla.org/en-US/docs/Web/API/WEBGL_compressed_texture_astc) is not available on almost any PC, so this is necessary for the foreseeable future
- BC textures are kept compressed when possible
- Rust is used for deswizzling the textures
  - Decoding BC6/BC7 (when needed) and ASTC textures is also done in Rust
- Rendering performance is in an ok state, but there's still room for improvement. The largest levels can approach 5ms frame time, while small and medium levels run at ≤ 2ms without any issues
- Animations only update at 30 FPS. This simulates the game better and is more efficient, but can result in some of them looking choppy
- Two new GfxFormats were added to support additional vertex attribute formats
 
image


magcius:
A couple of years ago, Jellonator tried to port the Tegra code and BC decoding over to Rust in 0c86931e072c55156eb46476dd533f477df20591 , but we had to revert it since it caused regressions. If you'd like to take a stab at re-landing that commit or converting more code back to Rust, please do so. (and please do it as a separate PR)

That said, the current tegra swizzle code is already in Rust <https://github.com/magcius/noclip.website/blob/main/rust/src/tegra_texture.rs>; I'd be happy to switch to the tegra_swizzle crate, but I'd prefer not to have two of them.

egartley:
> A couple of years ago, Jellonator tried to port the Tegra code and BC decoding over to Rust in [0c86931](https://github.com/magcius/noclip.website/commit/0c86931e072c55156eb46476dd533f477df20591) , but we had to revert it since it caused regressions. If you'd like to take a stab at re-landing that commit or converting more code back to Rust, please do so. (and please do it as a separate PR)
> 
> That said, the current tegra swizzle code is already in Rust https://github.com/magcius/noclip.website/blob/main/rust/src/tegra_texture.rs; I'd be happy to switch to the tegra_swizzle crate, but I'd prefer not to have two of them.

For porting the Tegra/BC code, I could try that out with the previous attempt, but I'm not very familiar with Rust.

So I did originally use the current Rust code for deswizzling, same as Odyssey and Mario Kart, but it doesn't return the expected amount of data that supports keeping BC textures compressed. Decoding afterwards works fine, though. This is some debug output for using the existing implementation, showing the texture name, format, mips count and byte length of each deswizzled mip level:

`tegra_deswizzle(texture.textureDataArray[0].mipBuffers[m].createTypedArray(Uint8Array), blockWidth, blockHeight, blockBytes, width, height, texture.blockHeightLog2) as Uint8Array<ArrayBuffer>`

image

And then when using the tegra_swizzle crate on the same level:

`pmtok_deswizzle(texture.textureDataArray[0].mipBuffers[m].createTypedArray(Uint8Array), width, height, blockWidth, blockHeight, blockBytes, 1) as Uint8Array<ArrayBuffer>`

image

I believe the discrepancy in sizes come from the block height log2/mip0 values and how tegra_swizzle handles it. If I modify my call to the crate's function to pass in `BRTI.blockHeightLog2` directly, instead of letting it infer the height, then it results in the same output as the existing implementation with the incorrect compressed sizes.

It makes sense not to have two different deswizzle functions, so I would argue for switching to the tegra_swizzle crate since I want to keep BC textures compressed when possible to help with load times. If that works, then I can remove the "pmtok" prefix and either replace the existing one or deprecate it. I could do that either in here or a separate request to test it in isolation first, whatever works best.

magcius:
Either/or is fine. I also wonder that since we already have texturedecoder2d, if we could replace the existing software BC decoder with it... I might try that myself though.

magcius:
Similarly, for ZSTD, impiaaa ported a bunch of our compression over from custom implementations to Rust in 38d94c41186a2ca376e2955c3ec5d54b09deac75, would be nice to keep that for zstd as well.

egartley:
The ZSTD decompression has been moved to Rust, so that's all set. I had to install LLVM in order for the build to be successful. Before that, it wasn't able to compile one of the dependencies, even with the Visual Studio C++ build tools installed.

For the deswizzling, I'm almost done but running into a weird issue, only in Mario Kart, where some mip levels aren't deswizzled correctly. Here's an example:

image

It only happens with BC1 formats with non-square dimensions, but some other BC1 non-square textures are perfectly fine. I couldn't find the same issue in either Origami King or Odyssey. This is the new contents of `tegra_texture.rs` currently:

```rust

use std::num::NonZeroU32;
use tegra_swizzle::surface::{BlockDim, deswizzle_surface, deswizzled_surface_size};
use wasm_bindgen::prelude::wasm_bindgen;

#[wasm_bindgen]
pub fn tegra_deswizzle(src: &[u8], width: u32, height: u32, block_width: u32, block_height: u32, bytes_per_block: u32, mipmap_count: Option<u32>, depth: Option<u32>, layer_count: Option<u32>, block_depth: Option<u32>) -> Vec<u8> {
    let block_dim: BlockDim = BlockDim {
        width: NonZeroU32::new(block_width).unwrap(),
        height: NonZeroU32::new(block_height).unwrap(),
        depth: NonZeroU32::new(block_depth.unwrap_or(1)).unwrap()
    };
    deswizzle_surface(width, height, depth.unwrap_or(1), src, block_dim, None, bytes_per_block, mipmap_count.unwrap_or(1), layer_count.unwrap_or(1)).unwrap()
}

#[wasm_bindgen]
pub fn tegra_deswizzled_size(width: u32, height: u32, block_width: u32, block_height: u32, bytes_per_block: u32, mipmap_count: Option<u32>, depth: Option<u32>, layer_count: Option<u32>, block_depth: Option<u32>) -> usize {
    let block_dim: BlockDim = BlockDim {
        width: NonZeroU32::new(block_width).unwrap(),
        height: NonZeroU32::new(block_height).unwrap(),
        depth: NonZeroU32::new(block_depth.unwrap_or(1)).unwrap()
    };
    return deswizzled_surface_size(width, height, depth.unwrap_or(1), block_dim, bytes_per_block, mipmap_count.unwrap_or(1), layer_count.unwrap_or(1));
}
```

I'll continue to debug over the weekend. It might be an issue with the crate or more likely just an edge case that needs to be handled differently.

egartley:
Figured it out and pushed up the changes. It was because when sizes weren't a multiple of 4, then the per-mip deswizzling wouldn't work with how tegra_swizzle handles block heights. I made a new function that deswizzles all mips at once then splits them back up, which fixed the problem. Mario Kart and Odyssey are working as expected now, but I don't have a copy of TMSFE so I can't test that one.

magcius:
You can set the DataFetcher to use data from the live website to test with.

  Jasper

egartley:
Ah ok, didn't think of that. Everything looks good now for the texture changes, MK8D, Odyssey and TMSFE are working the same as before.

magcius:
Mind sending me the data files for this game?

magcius:
Reverting this, as the addition of the zstd crate broke builds on Windows:

> warning: zstd-sys@2.0.16+zstd.1.5.7: Compiler family detection failed due to error: ToolNotFound: failed to find tool "clang": program not found (see https://docs.rs/cc/latest/cc/#compile-time-requirements for help)

egartley:
I was able to get it to work on Windows by installing LLVM. [https://github.com/llvm/llvm-project](https://github.com/llvm/llvm-project) I mentioned it before in one of the comments above. It was that same compile error. We could just use the original node module or whatever I had before for zstd instead of using rust.

PR #901, opened 2026-04-27
Wind Waker: disable glare when facing away from the sun

Nabile-Rahmani:
https://github.com/user-attachments/assets/af1fbb65-0888-4493-829c-8adc402af66d

https://github.com/user-attachments/assets/9557c82b-641c-4ece-8e0a-221bbe8ee10a

Z peeking (`numPointsVisible`) doesn't seem to consider the sun direction and darkens the environment when looking in the opposite direction.

I preemptively check the angle instead.

Do you find it a reasonable hack ? (I take it you've tried to match the decompiled code)

(Disclaimer: couldn't set up the dev env so I patched the production script for testing, it shouldn't make a difference though)

### DIFF
--- a/src/ZeldaWindWaker/d_kankyo_wether.ts
+++ b/src/ZeldaWindWaker/d_kankyo_wether.ts
@@ -1609,6 +1609,16 @@ function dKyr_sun_move(globals: dGlobals): void {
         // Sun should not glare during strange hours of night.
         sunCanGlare = false;
     }
+    else {
+        const sunForward = scratchVec3;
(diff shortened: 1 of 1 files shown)


magcius:
[The original code](https://github.com/zeldaret/tww/blob/main/src/d/d_kankyo_rain.cpp#L564-L630) doesn't do this as far as I can tell. That said, I can't seem to reproduce your issue, because the land / sea are read back from the Z buffer and are considered to "obscure" the sun. I wonder why that depth readback isn't working for you?

Nabile-Rahmani:
Oof, turns out it's a Firefox bug that was fixed in a more up to date package.

Sorry about that !

PR #903, opened 2026-05-01
Spyro: Add texture mips, more efficient rendering

egartley:
Added higher resolutions for textures and cleaned up a bunch of the rendering setup. Average performance gain is 0.1ms, so roughly 30% better. The shader is a lot simpler and vertex colors are used properly. Water also has a more accurate color.

I still plan to add mobys at some point, and I'll probably re-work the rendering setup again once that happens. But for now, it's in a pretty good state.

### DIFF
--- a/src/Spyro/bin.ts
+++ b/src/Spyro/bin.ts
@@ -14,7 +14,6 @@ interface LevelStream {
     vertices: number[];
     colors: number[];
     uvs: number[];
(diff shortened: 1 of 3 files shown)


magcius:
I assume this doesn't need new data?

egartley:
That's correct. Each level already has its entire subfile available, only just some of it is actually parsed for now.

PR #905, opened 2026-05-03
Add Ratchet & Clank

simon-paris:
This adds the first Ratchet & Clank game.

Most of the credit for the reverse engineering work goes to the [wrench](https://github.com/chaoticgd/wrench) developers.

0 veldin

#### Extracting files:

1. Place the NTSC version of the game iso at `data/RatchetAndClank1_raw/game.iso`
2. Run `pnpm build:RatchetAndClank 1` (there is a space between k and 1)
3. The files are output to `data/RatchetAndClank1`

#### What's supported

 * Tfrag meshes
   * Single-use world-space meshes for terrain
   * 3 lods, shares verts between lods
 * Tie meshes
   * Instanced static meshes
   * 3 lods, can morph between lods
 * Shrub meshes
   * Instanced static meshes
   * No lods, fades out
 * Skybox
   * Made of up to 8 sphere meshes with either vertex colors or a texture
   * Clear color is configurable
 * Collision mesh
   * World space mesh with 8 bit tags per face
   * Can be enabled in settings
 * LOD transitions
   * Supports morphing ties between lods
   * Supports fading out shrubs (doesn't support billboards)
   * Lod preset and bias can be configured in settings
 * Moby positions and paths
   * Can be enabled in settings

#### What's not supported

 * Moby meshes
 * Most liquid surfaces, ice surfaces and some glass surfaces
 * Proper transparency ordering
 * Shrub billboards
 * Tie lod morphing affects position only, should also morph normal and color
 * Texture viewer (because I use array textures)

#### Changes to shared code

 * Support instancing unindexed meshes (ties and shrubs don't have indices)

#### Performance

 * On my machine, renders in ~0.5ms - 2ms and 75-250 draw calls, depending on the level
 * Takes ~500ms to parse level files
 * Level files are ~10-14MB, they're extracted with the original compression as-is

#### Render options

<details>
<summary>Show</summary>
settings
</details>

#### More Screenshots

<details>
<summary>Show</summary>
1 novalis
2 aridia
3 kerwan
4 eudora
5 rilgar
6 nebula g34
7 umbris
8 batalia
9 gaspar
10 orxon
11 pokitaru
12 hoven
13 oltanis orbit
14 oltanis
15 quartu
16 kalebo iii
17 veldin orbit
18 veldin
13 oltanis collision
</details>


src/RatchetAndClank/render-shrub.ts
```diff
@@ -0,0 +1,342 @@
...
+                const farDist = shrubInstance.drawDistance + settingLodBias * 1.5;
+                if (farDist > 0) {
+                    const nearDist = farDist * 0.5;
+                    lodAlpha = 1 - (distanceToCamera - nearDist) / (farDist - nearDist);
+                    lodAlpha = Math.max(0, Math.min(1, lodAlpha));
+                }
+            }
+            if (lodAlpha <= 0) continue;
+
+            // this is much slower than doing nothing because of the jumps into rust
```
magcius, line 341: What browser were you testing on? When I checked a little while ago, I found that the overhead going into wasm and back (especially with no memory transfer) was basically free on Chrome.

src/RatchetAndClank/bin-core.ts
```diff
@@ -0,0 +1,1575 @@
+import { IS_DEVELOPMENT } from "../BuildVersion";
+import { GsPrimitiveType } from "../Common/PS2/GS";
+import { DataViewExt } from "./DataViewExt";
+import { assert } from "../util";
+import { getBits, ImaginaryGsCommand, ImaginaryGsCommandBuffer, truncateTrailing0xFF } from "./utils";
+import { readVifCommandList, VifUnpackFormat, VifUnpackReader } from "./vif";
+
+export type GsRamTableEntry = {
```
magcius: In general I've preferred TypeScript's `interface` over `type` because it has [much faster type checking performance](https://github.com/microsoft/TypeScript/wiki/Performance#preferring-interfaces-over-intersections).

src/RatchetAndClank/render-tie.ts
```diff
@@ -0,0 +1,409 @@
...
+            vertexArrayBuffer[ptr++] = positionScale * vert.lodMorphOffsetY;
+            vertexArrayBuffer[ptr++] = positionScale * vert.lodMorphOffsetZ;
+        }
+    }
+
+    for (const strip of strips) {
+        for (let i = 0; i < strip.verts.length - 2; i++) {
+            pushTriangle([strip.verts[i + 0], strip.verts[i + 1], strip.verts[i + 2]], strip.material.texture, strip.material.clamp);
+        }
+    }
```
magcius: Why not use an index buffer here? Because of fixTexcoords?

src/RatchetAndClank/shader-lib.ts
```diff
@@ -0,0 +1,244 @@
...
+float fogFactor() {
+    float worldDepth = linearizeDepth(1.0 - gl_FragCoord.z, u_NearFarClip.x, u_NearFarClip.y);
+    float fogFactor = 1.0 - clamp((u_FogParams.farDist - worldDepth) / (u_FogParams.farDist - u_FogParams.nearDist), 0.0, 1.0);
+    fogFactor = u_FogParams.nearIntensity + fogFactor * (u_FogParams.farIntensity - u_FogParams.nearIntensity);
+    return fogFactor;
+}
+
+vec3 adjustSaturation(vec3 color, float adjustment) {
+    const vec3 luminanceWeights = vec3(0.2125, 0.7154, 0.0721);
+    float luminance = dot(color, luminanceWeights);
```
magcius: MonochromeNTSCLinear

src/RatchetAndClank/shader-lib.ts
```diff
@@ -0,0 +1,244 @@
...
+
+const float SATURATION_ADJUST = 1.15;
+
+float linearizeDepth(float depth, float near, float far) {
+    float z = depth * 2.0 - 1.0;
+    return (2.0 * near * far) / (far + near - z * (far - near));
+}
+
+float fogFactor() {
+    float worldDepth = linearizeDepth(1.0 - gl_FragCoord.z, u_NearFarClip.x, u_NearFarClip.y);
```
magcius: This won't work on orthographic projections, and is overly complicated. If spherical depth won't work for you (`length(worldPos - cameraPos)`), then you could pass view-space Z from the vertex shader.

(Alternatively, `1.0 / gl_FragCoord.w` might have what you want? I forget...)

src/RatchetAndClank/shader-lib.ts
```diff
@@ -0,0 +1,244 @@
...
+}
+
+vec4 commonVertexLighting(vec4 rgba, vec3 normal, vec4 dirLightIndices) {
+    vec4 light = rgba;
+
+    // directional
+    int lightCount = 0;
+    for(int i = 0; i < 4; i++) {
+        int dirLightIndex = int(dirLightIndices[i]);
+        if (isNullLight(i, dirLightIndex)) lightCount++;
```
magcius: You don't use lightCount here?

src/RatchetAndClank/shader-lib.ts
```diff
@@ -0,0 +1,244 @@
...
+}
+
+vec4 applyDirectionalLight(vec3 normal, int dirLightIndex) {
+    DirectionLight dirlight = u_DirectionLights[dirLightIndex];
+
+    const vec4 NEGATIVE_ALPHA = vec4(1.0, 1.0, 1.0, -1.0);
+
+    vec4 light = vec4(0.0);
+    float nDotL_A = dot(normal, dirlight.directionA);
+    if (nDotL_A > 0.0) light += nDotL_A * dirlight.colorA * NEGATIVE_ALPHA;
```
magcius: You could adjust NEGATIVE_ALPHA when packing into colorA/B.

src/RatchetAndClank/render-sky.ts
```diff
@@ -0,0 +1,304 @@
...
+layout(location = ${SkyProgram.a_ST}) in vec2 a_ST;
+layout(location = ${SkyProgram.a_Rgba}) in vec4 a_Rgba;
+layout(location = ${SkyProgram.a_Alpha}) in float a_Alpha;
+
+out vec2 v_ST;
+out vec4 v_Rgba;
+
+void main() {
+    vec4 t_PositionWorld = UnpackMatrix(u_SkyTransform) * vec4(a_Position.xyz, 1.0f);
+    gl_Position = (UnpackMatrix(u_ClipFromWorld) * t_PositionWorld).xyww; // infinite depth
```
magcius: This won't work on orthographic views.

src/RatchetAndClank/render-sky.ts
```diff
@@ -0,0 +1,304 @@
...
+
+    float isTextured = u_ExtraData.x;
+    if (isTextured == 1.0) {
+        v_ST = a_ST;
+        v_Rgba = vec4(1.0, 1.0, 1.0, a_Alpha);
+    } else {
+        v_Rgba = vec4(a_Rgba.rgb, a_Alpha * a_Rgba.a);
+    }
```
magcius: Why not merge Rgba and Alpha at vertex load time?

src/RatchetAndClank/render-tie.ts
```diff
@@ -0,0 +1,409 @@
...
+    renderTie(renderInstList: GfxRenderInstList, tieGeometriesByLod: (TieGeometry | null)[], tieClass: TieClass, tieInstanceBatch: TieInstance[], textureMappings: GfxSamplerBinding[], cameraPosition: vec3, settingLodPreset: number, settingLodBias: number, instanceDataBuffer: MegaBuffer): void {
+        const scratchVec3 = vec3.create();
+
+        type TieDrawInstance = { objectMatrix: mat4, directionLights: number[], rgbasRow: number, lodMorphFactor: number, i: number };
+        const tieInstancesToDrawByLod: TieDrawInstance[][] = [[], [], []];
+        for (let i = 0; i < tieInstanceBatch.length; i++) {
+            const tieInstance = tieInstanceBatch[i];
+
+            // tie instance transform
+            const objectMatrix = mat4.create();
```
magcius: My profiling has shown that creating new Float32Array's still has a non-zero construction cost. I'd bake this transform into the tieInstance matrix.

src/RatchetAndClank/render-shrub.ts
```diff
@@ -0,0 +1,342 @@
...
+
+layout(location = ${ShrubProgram.a_Position}) in vec3 a_Position;
+layout(location = ${ShrubProgram.a_Normal}) in vec3 a_Normal;
+layout(location = ${ShrubProgram.a_TextureParams}) in vec2 a_TextureParams; // x = texture index, y = clamp flag
+layout(location = ${ShrubProgram.a_ST}) in vec2 a_ST;
+
+layout(location = ${ShrubProgram.a_InstanceTransform0}) in vec4 a_InstanceTransform0;
+layout(location = ${ShrubProgram.a_InstanceTransform1}) in vec4 a_InstanceTransform1;
+layout(location = ${ShrubProgram.a_InstanceTransform2}) in vec4 a_InstanceTransform2;
+layout(location = ${ShrubProgram.a_InstanceTransform3}) in vec4 a_InstanceTransform3;
```
magcius, line 64: Can't wait until we have real buffers (alternatively, do the texture packing trick we do in other places)

src/RatchetAndClank/utils.ts
```diff
@@ -0,0 +1,214 @@
...
+     */
+    upload: () => void,
+
+    destroy: () => void,
+};
+
+/**
+ * Create a shared buffer for uploading instance data.
+ */
+export function createMegaBuffer(device: GfxDevice, name: string, initialSizeInBytes: number): MegaBuffer {
```
magcius, line 109: Did you look at `dynamicBufferCache` at all?

magcius (commented):
Initial set of feedback. None of this has to be addressed.

In general I notice you aren't using gl-matrix vector types frequently, and aren't using standard utilities like Color. That is fine, but it does mean that this code is a bit different convention-wise from other noclip projects.

You also seem to pass around a lot of objects rather than baking more things up-front. I'm worried about the perf cost of something like that.

simon-paris:
Thanks for the feedback. Might take a few days to work on it, I've got a pretty busy week.

src/RatchetAndClank/shader-lib.ts
```diff
@@ -0,0 +1,251 @@
...
+
+float linear01Depth() {
+    float depth = gl_FragCoord.z;
+    float near = u_CameraData.near;
+    float far = u_CameraData.far;
+
+    float z = (1.0 - gl_FragCoord.z) * 2.0 - 1.0;
+    float depthWorld = (2.0 * near * far) / (far + near - z * (far - near));
+    float depth01 = (depthWorld - near) / (far - near);
+    return depth01;
```
magcius: This formulation currently won't work for orthographic matrices, nor will it work for WebGPU (because of the different clip space convention). Rather than do this, I think it's easier to just multiply by the inverse-2x2 of the view-projection matrix, which should handle everything generically.

<https://github.com/magcius/noclip.website/blob/main/src/DarkSouls/render.ts#L2134-L2140>
<https://github.com/magcius/noclip.website/blob/main/src/DarkSouls/render.ts#L2212-L2223>

src/RatchetAndClank/shader-lib.ts
```diff
@@ -0,0 +1,251 @@
...
+
+    if (rgba.a >= 1.0 && light.a < 1.0) {
+        light.a = rgba.a;
+    }
+
+    return light;
+}
+
+float fogFactor(vec3 vertexPosWorld) {
+    if (u_CameraData.isOrtho == 1.0) return 0.0;
```
magcius: Any reason to turn off fog on ortho?

src/RatchetAndClank/shader-lib.ts
```diff
@@ -0,0 +1,251 @@
...
+        4, // sky color
+        8, // fog params
+        16 * 16, // directional lights
+        4 * 256 * 3, // texture remaps (3 arrays of 256 vec4s)
+    ].reduce((a, b) => a + b, 0),
+    SceneParams: `
+
+// size 12
+struct CameraData {
+    vec3 position;
```
magcius: I generally don't trust vec3 or raw floats in GLSL std140 packing, especially because the size of the struct will change when used in an array of them. I'd use vec4's for all of these.

src/RatchetAndClank/bin-core.ts
simon-paris: Fixed

src/RatchetAndClank/render-shrub.ts
simon-paris: I'm on chrome on windows.

It's pretty dramatically slower, and even slower again if I have the devtools open. I'll test it on other browsers and mac too.

src/RatchetAndClank/render-tie.ts
simon-paris: Fixed

src/RatchetAndClank/shader-lib.ts
simon-paris: Fixed

src/RatchetAndClank/shader-lib.ts
simon-paris: Fixed

src/RatchetAndClank/render-tie.ts
simon-paris: Yes, because of the texcoord wrapping. I don't think I can emulate that in the shader without some nasty trickery

src/RatchetAndClank/shader-lib.ts
simon-paris: Fixed

src/RatchetAndClank/shader-lib.ts
simon-paris: I've changed the fog to use camera distance and moved it to the vertex shader (that's more accurate too, the ps2 takes fog intensity per-vertex)

I've changed the mip selection to use the same math as the ps2, which is 1/gl_FragCoord.w.

src/RatchetAndClank/utils.ts
simon-paris: I'll try changing to that

src/RatchetAndClank/render-sky.ts
simon-paris: Fixed

src/RatchetAndClank/render-sky.ts
simon-paris: Fixed

src/RatchetAndClank/shader-lib.ts
simon-paris: Fixed

src/RatchetAndClank/shader-lib.ts
simon-paris: Well the distance calculation won't work since the camera is so far away... I suppose I'll just choose some value, like `(nearIntensity + farIntensity) / 2.0`

src/RatchetAndClank/shader-lib.ts
simon-paris: I've changed mip selection to use 1/w, which is how the manual says it should work

Image

src/RatchetAndClank/render-tie.ts
magcius: Interesting, I figured the manual-bilinear texture windowing stuff you already do in the pixel shader could emulate this. But that's fine.

src/RatchetAndClank/render-shrub.ts
magcius: I've also noticed V8 performance to be flakey and have strange regressions with newer Chrome versions, I wonder if it's regressed since I tested. I'll consider this OK for now and put it on my TODO list to re-test wasm culling again. Sigh.

magcius:
1. Can you make sure that you're gfx leak-free? Set `main.loadSceneDelta = 0` in your F12 console, then press F9 on any given scene to reload it. If you're leaking gfx resources, you should get a warning in your browser console.

2. Mind sending me the data files, either through Discord DMs or to the email address in my profile.

src/RatchetAndClank/render-shrub.ts
simon-paris: My quick test results:
 * Chrome windows:  1ms without culling, 3ms with culling, 15ms with culling if devtools open
 * Chrome mac: about the same
 * Firefox windows: 1.5ms without culling, 3ms with culling

simon-paris:
> Can you make sure that you're gfx leak-free? Set main.loadSceneDelta = 0 in your F12 console, then press F9 on any given scene to reload it. If you're leaking gfx resources, you should get a warning in your browser console.

Verified leak free

chaoticgd:
Just noticed Ratchet & Clank got added to noclip.website. Really cool!

PR #909, opened 2026-05-20
Add Rumble Racing (PS2)

mattbruv:
Well, this was a marathon of a coding project... This PR adds Rumble Racing to noclip.

A few noteworthy things about this implementation:

1. The game calculates the actors' Y level, scale, and rotations at runtime using lots of heuristics based on the map's terrain, actor properties, etc. etc.. To get around having to spend another half year of my prime years in life reverse engineering all of this logic just to move a chicken up 0.1 meter (this is either genius or lazy, you decide) I took a save state using PCSX2 at the exact moment after loading the level, and scraped out the final actor Y levels and transformations. I've saved all of these in a JSON file based on the map and actor ID saved in the DATA folder. All of this data simply places the instanced actor models where they should be, and works for like 98% of the actors.
2. This is my first experience with rendering 3D geometry (outside of previously writing this data to a gltf file), so I have probably made some horribly naive decisions as far as rendering the scene. Please let me know if I'm doing something dumb.

Known Visual issues:
1. I was never able to figure out the alternating unwinding face issue when manually unwinding triangle strips, so I don't attempt any back-face culling
2. I have done no reverse engineering of how the game does lighting, so the scenes have very basic lighting. Luckily the maps are still fun and visually OK to look at despite this issue
5. I have not looked into how the game creates the panorama / skybox.
6. I do not handle some alpha/transparent textures/geometry correctly yet. While imperfect, it doesn't take away from things too much.

image


mattbruv:
A few more things I'd like to add when thinking more about this PR. Since the AI code in the other PR is getting torn up, I also want to be transparent up front about my AI usage before you review it so you don't go crazy wondering

First of all, please don't be dissuaded by reviewing this PR based on the 43,000 lines added. Almost 90% of that (37,000 lines) is just JSON data for actor transformations. Ignoring that, the actual amount of relevant LOC is like 5,000 LOC. I am definitely open to storing the JSON data in a different format/approach which would be more compressed and/or not inflate file size/LOC.

As far as how AI was used, I used Gemini/Claude to:
- write decompress.go. The decompression routine was written entirely by AI
- A fair amount of the Go code have been iterated on to varying degrees during the RE process, between both me and AI. In the Go codebase, I used AI as a tool to write out repetitive code in scenarios where it would have been annoying, (a majority of the VIF command parsing and VIF debug code for example).
- The noclip specific TS code/structure is almost entirely AI written, and the shader code is 100% AI written). I basically started with the Example scene, and iterated on it with AI to start drawing simple models, then actors, and so on, until we're at the state it is in now.

There are lots of comments in the Go code which are remnants of earlier code, debugging code commented out, and so on. Since I basically copied the relevant code from my CLI tool and pasted it here, a fair amount of the comments and some code could be considered irrelevant. 

As we briefly discussed a while ago, you guys said you'd be open to adding Go as a language to noclip. If, after looking at this PR, it seems like it would introduce more annoyances than it's worth, I am fine with re-writing the parsing code in TS / Rust in order to introduce less friction. If you'd rather have me scrap the Go stuff, it won't hurt my feelings.

magcius:
Thanks for the heads up. Sorry I'm so backed up with reviews.

> Almost 90% of that (37,000 lines) is just JSON data for actor transformations. Ignoring that, the actual amount of relevant LOC is like 5,000 LOC. I am definitely open to storing the JSON data in a different format/approach which would be more compressed and/or not inflate file size/LOC.

For data like this, rather than have it be part of the initial code bundle, we could instead store it along with the game data, fetched at runtime.

mattbruv:
@magcius, I was able to have the all of the Go code ported to TS. Now everything is purely in TS. After some reflection I felt like asking noclip to support Go just for one game would unnecessarily introduce complications and bloat. Amazingly, Claude Sonnet 4.6 was able to one-shot the port and keep the code faithful to my original code. I've tested it locally and everything is good, it even feels like it loads faster now than before.

src/RumbleRacing/helpers/bytes.ts
```diff
@@ -0,0 +1,110 @@
...
+
+export function readUint32LE(data: Uint8Array, offset: number): number {
+  return new DataView(data.buffer, data.byteOffset, data.byteLength).getUint32(
+    offset,
+    true,
+  );
+}
+
+export function readUint16LE(data: Uint8Array, offset: number): number {
+  return new DataView(data.buffer, data.byteOffset, data.byteLength).getUint16(
```
magcius: Please use DataView directly; these wrappers add very little.

src/RumbleRacing/asset/o3d/chunk.ts
```diff
@@ -0,0 +1,124 @@
...
+    if (chunkEnd > data.length) {
+      throw new Error(
+        `chunk size ${size} exceeds remaining data at offset ${offset}`,
+      );
+    }
+    const payload = data.slice(offset, chunkEnd);
+
+    const assetChunk: AssetChunk = {
+      offset,
+      magic: new TextEncoder().encode(magic),
```
magcius: ? Why do you decode and then encode?

src/RumbleRacing/asset/o3d/chunk.ts
```diff
@@ -0,0 +1,124 @@
...
+      case 2:
+        if (magic !== "ELDA") throw new Error("NOT AN ELDA! Got: " + magic);
+        currentElda = parseELDA(assetChunk);
+        break;
+    }
+
+    offset = chunkEnd;
+    chunkIndex++;
+
+    if (currentElhe && currentEltl && currentElda) {
```
magcius, line 120: Does this condition ever happen? If not, please assert on it.

src/RumbleRacing/asset/o3d/vif.ts
```diff
@@ -0,0 +1,273 @@
+import { readUint32LE, uint32ToFloat32 } from "../../helpers/bytes";
+
+export type VifCommandKind =
```
magcius: Please use the common enums added in ee4d72f92c7987c17450d21f1b9c53acc79745de rather than these strings.

src/RumbleRacing/helpers/pstm8.ts
```diff
@@ -0,0 +1,62 @@
...
+  const result = new Array<T>(256);
+  for (let i = 0; i < 256; i++) {
+    const coord = mapLinearIndexToCoord(i);
+    const flatIndex = coord.y * 16 + coord.x;
+    result[flatIndex] = data[i];
+  }
+  return result;
+}
+
+export function swizzleClutPstm4_16<T>(data: T[]): T[] {
```
magcius: I believe the existing `gsMemoryMapReadImagePSMT4_PSMCT32` will do what you want?

src/RumbleRacing/asset/o3d/geometry.ts
```diff
@@ -0,0 +1,222 @@
+import { VifCommand, UnpackData } from "./vif";
+
+export interface Vertex {
```
magcius: Please use vec3/vec2 instead of this.

src/RumbleRacing/chunk/shoc/decompress.ts
```diff
@@ -0,0 +1,140 @@
+export function decompress(src: Uint8Array, outSize: number): Uint8Array {
+  let i = 0;
+  let dst: number[] = [];
+  const n = src.length;
+
+  const readU8 = (idx: number): number => {
+    if (idx < 0 || idx >= n) return 0;
```
magcius: Does this condition happen?

src/RumbleRacing/chunk/shoc/shoc.ts
```diff
@@ -0,0 +1,49 @@
...
+export interface Shoc {
+  kind: "SHOC";
+  index: number;
+  startAddress: number;
+  metadata: ShocMetadata;
+  data: Uint8Array;
+}
+
+function parseSubChunk(data: Uint8Array, shocIndex: number): ShocMetadata {
+  const fourCCbytes = data.slice(8, 12);
```
magcius: readString()?

magcius (changes requested):
Initial review.

src/RumbleRacing/asset/o3d/chunk.ts
mattbruv: This condition is only expected to succeed every third iteration. In practice the chunks will always be in groups of three (Will always be an `ELHE` followed by an `ELTL` followed by an `ELDA`). We don't want to group them into their own object together until all three have been parsed out. The if statement is more checking to see if we've parsed enough data yet, rather than asserting that they are there, although it does expect them to be there on the third iteration.

src/RumbleRacing/chunk/shoc/decompress.ts
mattbruv: Good eye. I tested it just now and this condition never triggers in the retail or prototype data, so I've removed this check.

src/RumbleRacing/helpers/pstm8.ts
mattbruv: Most likely. I've gone ahead and referenced noclip's existing PS2 GS enums. Figuring out how PS2 texture swizzling was confusing for me at the time when I was writing the original Go code. I was able to figure out how swizzling and CLUTs worked on my own and implemented it in the way that made sense to me, but it is not ideal to have multiple implementations. I'll have to refresh myself on how this all works again and see if I can utilize noclip's implementation

mattbruv:
Alright, I got the lighting and skybox figured out, and it looks so much better!!! Hopefully this will be the last big code change I drop on you in this PR because it is much higher fidelity now. I am really happy with the state it is in.

My latest commit adds:
- Vertex RGBA colors (which solves the lighting and skybox color issues)
- Translucent geometry is now rendering properly (windows, glass panes, water, clouds, etc.).
- Huge performance gains by batching geometry and draw calls. (went from ~60 FPS on some larger maps to now every map running at ~1000+fps)
- Added mipmap levels to the textures (mip maps were able to be parsed out previously, but I was only using highest resolution texture)
- Added a separate pane for toggling various geometry, and added a toggle for textures/vertex colors. (this is kinda cool because if you disable textures you can see how the vertex coloring is affecting the scene).
- Improved some default save states since the rendering is much prettier now.

I did use Claude to generate a fair amount of the changes. I have already gone through this new code in its entirety in a separate PR on my own and have made it use existing noclip mannerisms and cleaned it up to my taste (as much as I know about shaders). I'm going to mark it as ready to be reviewed again while I work on your PS2 swizzling comment from earlier.

image
image
image

src/RumbleRacing/helpers/pstm8.ts
mattbruv: Alright, I added `gsMemoryMapReadImagePSMT4_PSMCT16` and `gsMemoryMapReadImagePSMT8_PSMCT16` and extracted some existing re-usable logic into a function `convertPixelPSMCT16` which is used by all three functions now, and got rid of my custom swizzling code to now use these functions. Not going to mark this as resolved until you give the OK for those changes.

mattbruv:
Alright, I've added support for instanced lights and am now rendering the powerup models/glows/animations.

This is the last thing I'm going to add to this PR outside of whatever you request me to change. I could keep adding improvements/features but I want to add any new content to a later PR to not keep inundating you with more code to review in the same initial PR.

https://github.com/user-attachments/assets/327b97ba-9d42-441e-ac63-1016a8c9cff0

mattbruv:
Hosting it [here](https://mattbruv.github.io/noclip.website/) for anyone who wants to check it out

src/RumbleRacing/Scenes.ts
```diff
@@ -0,0 +1,918 @@
...
+    "Beach Blast",
+    new RumbleRacingSceneDesc("BB1", "SunBurn", "Sun Burn", {
+      textureIds: new Set([3120]), // tornado clouds
+    }),
+    new RumbleRacingSceneDesc("BB2", "SurfAndTurf", "Surf And Turf", {
+      textureIds: new Set([3120]), // tornado clouds
+    }),
+    "Bad Lands",
+    new RumbleRacingSceneDesc("BL1", "SoRefined", "So Refined", {
+      textureIds: new Set([1584]), // tornado clouds
```
PFedak: are there just two different tornado cloud textures?

src/RumbleRacing/Scenes.ts
```diff
@@ -0,0 +1,918 @@
...
+    }),
+    "Southern Exposure",
+    new RumbleRacingSceneDesc("SE1", "TrueGrits", "True Grits", {
+      textureIds: new Set([3120]), // tornado clouds
+    }),
+    new RumbleRacingSceneDesc("SE2", "OverEasy", "Over Easy", {
+      textureIds: new Set([3120]), // tornado clouds
+    }),
+  ],
+  hidden: !IS_DEVELOPMENT,
```
PFedak:
```suggestion
```

src/RumbleRacing/rumbleRacing.ts
```diff
@@ -0,0 +1,280 @@
...
+          positions.push(vert.position);
+          normals.push(vert.normal ?? vec3.fromValues(0, 0, 0));
+          colors.push(vert.color ?? White);
+          uvs.push(vert.uv);
+          if (vert.color !== null) hasVertexColors = true;
+        }
+
+        let isFlipped = false;
+        for (let i = 2; i < strip.vertices.length; i++) {
+          if (strip.vertices[i].adcBitSet) {
```
PFedak: maybe worth a comment here, just summarizing our discord convo lol

src/RumbleRacing/Scenes.ts
```diff
@@ -0,0 +1,918 @@
...
+  GlowBatch,
+  GlowDef,
+  GlowRenderer,
+  GlowShape,
+  glowColorFromRGBA32,
+} from "./Glow";
+import { buildGlowDefs } from "./Lights";
+
+const pathBase = `RumbleRacing`;
+const GLOBAL_SCALE = 300.0; // this feels the best
```
PFedak: you can also adjust `sceneMoveSpeedMult` and the near/far plane distances to get more control over the "feel" of things without messing with the underlying positions

PFedak:
Playing around a bit, is the Wild Kingdom skybox right? I haven't dug into what's going on but it looks pure black, despite having some blueish vertex colors, and there's odd z-fighting or bad sampling behavior visible if you have just textures visible.

src/RumbleRacing/Scenes.ts
```diff
@@ -0,0 +1,918 @@
...
+    const shared =
+      await sceneContext.dataShare.ensureObject<RumbleRacingShared>(
+        `${pathBase}/shared`,
+        async () => {
+          const data = await sceneContext.dataFetcher.fetchData(
+            `${pathBase}/DATA/GLBLDATA.TRK`,
+          );
+
+          const globalData: RumbleRacingTrackFile = processTrackFile(
+            new Uint8Array(data.arrayBuffer),
```
PFedak: It would touch a lot of files but I think the parsing code would be improved if you left the data as an `ArrayBufferSlice` - it has a lot of convenience features so you can e.g. create slices at different offsets without copying and easily get typed arrays and data views corresponding to those offsets.

src/RumbleRacing/rumbleRacing.ts
```diff
@@ -0,0 +1,280 @@
...
+export interface BoundingSphere {
+  center: vec3;
+  radius: number;
+}
+
+export const enum ActorType {
+  PowerUp = 8,
+}
+
+type MatrixRow = [x: number, y: number, z: number, w: number];
```
PFedak: I didn't actually know you could define an array like this - the names are just for like documentation purposes here, right? As far as I understand, your underlying JSON would be compatible with `type ActorMatrix = number[][]`. Do you know if the parser will validate the length with it defined like this

src/RumbleRacing/asset/o3d/vif.ts
```diff
@@ -0,0 +1,243 @@
...
+
+export interface UnpackData {
+  type: VifUnpackFormat;
+  v4_32: V4_32Entry[];
+  v3_32: V3_32Entry[];
+  v2_32: V2_32Entry[];
+  v4_8: V4_8Entry[];
+}
+
+export interface VifCommand {
```
PFedak: could also do this as a tagged union but this is fine

src/RumbleRacing/asset/o3d/vif.ts
```diff
@@ -0,0 +1,243 @@
...
+          const count = num;
+          const unpack: UnpackData = {
+            type: info.unpackFormat,
+            v4_32: [],
+            v3_32: [],
+            v2_32: [],
+            v4_8: [],
+          };
+
+          switch (info.unpackFormat) {
```
PFedak: can you add a `default` that throws an exception in this switch?

src/RumbleRacing/asset/o3d/o3d.ts
```diff
@@ -0,0 +1,96 @@
...
+function readVec3(view: DataView, offset: number): vec3 {
+  return vec3.fromValues(
+    view.getFloat32(offset + 0x0, true),
+    view.getFloat32(offset + 0x4, true),
+    view.getFloat32(offset + 0x8, true),
+  );
+}
+
+function parseGmdBounds(data: Uint8Array): GmdBounds | null {
+  if (data.byteLength < GMD_SPHERE_RADIUS + 4) return null;
```
PFedak: ha, this line threw me since I didn't realize it was an offset. What happens when this gets hit, like a bounding box without the sphere? or just empty?

feels a little odd to just define offset constants here and use raw numbers elsewhere, although I guess most of the others are consecutive

src/RumbleRacing/asset/asset.ts
```diff
@@ -0,0 +1,26 @@
+import { SHDR } from "../chunk/shoc/shdr";
+
+export interface Asset {
```
PFedak: I don't really understand this interface. Why are these methods? It seems like this is only used to print the chunk type, and nothing else extends it. 

I think having a catchall for debugging is good though

mattbruv:
> Playing around a bit, is the Wild Kingdom skybox right? I haven't dug into what's going on but it looks pure black, despite having some blueish vertex colors, and there's odd z-fighting or bad sampling behavior visible if you have just textures visible.

Hmmm.. I'm not getting an Z-fighting or any weird behavior for that level when disabling everything but textures. Could you show me a video of what you're seeing? The game does dynamically generate stars in the sky for some night levels like Wild Kingdom, Car Go, and Touch and Go (I believe) and I haven't gotten around to RE'ing that yet, I put it in the included README as a potential improvement, but if you think it will help the fidelity on this level I can look into it

PFedak:
> Hmmm.. I'm not getting an Z-fighting or any weird behavior for that level when disabling everything but textures.
huh maybe it's something weird with my setup, I'll dig in a little more

to be clear neither fixing this (assuming it's actually a problem lol) nor something like the star thing is a requirement for merging, just something I noticed

PR #921, opened 2026-06-16
Normalized Camera Speed

AdamTadeusz:
## Changes
- Normalized fpscamera speed. Holding multiple movement keys now doesn't accelerate the camera anymore than when holding a single movement key. Does not take into account moving up and down the y axis when useViewUp == true
- Early return from fpscamera update function when holding the camera reset button (b)
- Show camera speed in the statistics page

## TODO

- [x] remove importantUpdateDue. Some logic issue causes the url to be spammed without it, not present on main, fix.
- [x] ~~drag now applied in all directions always, could have negative impact on phone controls, if so revert.~~ Reverted the change, as it was impacting importantUpdate

## Pull Request Checklist

### DIFF
--- a/src/Camera.ts
+++ b/src/Camera.ts
@@ -227,8 +227,8 @@ export class FPSCameraController implements CameraController {
     private mouseMovement = vec3.create();
 
     private keyMoveSpeed = 60;
-    private keyMoveShiftMult = 5;
-    private keyMoveSlashMult = 0.1;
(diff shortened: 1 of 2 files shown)


magcius:
I think I see the current movement as a feature, not a bug. The code complexity that comes from normalizing the different movement sources this also doesn't seem worth it to me right now.

PR #924, opened 2026-06-22
Ratchet & Clank - Fix moby textures on mobile, fix tfrag transparency

simon-paris:
* Packing the texture remap data into floats wasn't working on Android. I've moved the final texture ids into the vertex buffer so no mapping in the shader is required.
 * Improve transparency ordering in tfrags. Split opaque and transparent, render transparent twice, once with depth write+alpha test, and again without. This emulates the ps2's TEST_1.AFAIL=FB_ONLY behavior. Mostly this fixes the artifacts on vines and grass hanging off ledges.
 * Small progress on rac2

## Pull Request Checklist

### DIFF
--- a/src/RatchetAndClank/render-tfrag.ts
+++ b/src/RatchetAndClank/render-tfrag.ts
@@ -1,16 +1,16 @@
 import { createBufferFromData } from "../gfx/helpers/BufferHelpers";
 import { GfxShaderLibrary } from "../gfx/helpers/GfxShaderLibrary";
-import { GfxBuffer, GfxBufferFrequencyHint, GfxBufferUsage, GfxDevice, GfxFormat, GfxInputLayout, GfxProgram, GfxSamplerBinding, GfxSamplerFormatKind, GfxTextureDimension, GfxVertexBufferFrequency } from "../gfx/platform/GfxPlatform";
+import { GfxBuffer, GfxBufferFrequencyHint, GfxBufferUsage, GfxCompareMode, GfxDevice, GfxFormat, GfxInputLayout, GfxProgram, GfxSamplerBinding, GfxSamplerFormatKind, GfxTextureDimension, GfxVertexBufferFrequency } from "../gfx/platform/GfxPlatform";
 import { GfxRenderCache } from "../gfx/render/GfxRenderCache";
 import { DeviceProgram } from "../Program";
 import { assert } from "../util";
 import { RatchetShaderLib } from "./shader-lib";
 import { Tfrag, TfragLight, TfragVertexInfo } from "./bin-core";
-import { PaletteTexture } from "./textures";
+import { packRemap, PaletteTexture, TextureAtlases } from "./textures";
 import { GfxRenderHelper } from "../gfx/render/GfxRenderHelper";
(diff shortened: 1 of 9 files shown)


src/RatchetAndClank/render-tfrag.ts
```diff
@@ -335,23 +389,40 @@ export class TfragRenderer {
...
+
+        for (let i = 0; i < draws.length; i++) {
+            const draw = draws[i];
+
+            const vertexCount = draw.transparentGeometry ? vertexData.vertexCountTransparent : vertexData.vertexCountOpaque;
+            if (!vertexCount) continue;
+            const vertexBuffer = draw.transparentGeometry ? vertexData.vertexBufferTransparent : vertexData.vertexBufferOpaque;
+
+            const renderInst = this.renderHelper.renderInstManager.newRenderInst();
+            renderInst.setMegaStateFlags({ depthWrite: draw.depthWrite, depthCompare: GfxCompareMode.Greater });
```
magcius, line 397: I don't quite understand how the second transparent draw is doing anything when you have the compareMode set to Greater; the second transparent draw will have the same positions as before, thus all the depth tests will fail. Should the second transparent draw be Equal?

I also don't see any alpha blending, so I'm not sure how this will make a difference in the final rendered image.

src/RatchetAndClank/render-tfrag.ts
```diff
@@ -128,12 +131,17 @@ export class TfragGeometry {
...
+            const vertexDataTransparent = this.assemble(lod, true);
+            const vertexBufferTransparent = createBufferFromData(device, GfxBufferUsage.Vertex, GfxBufferFrequencyHint.Static, vertexDataTransparent.vertexArrayBuffer.buffer);
+            device.setResourceName(vertexBufferTransparent, `Tfrag LOD ${lod} Transparent VB`);
             this.lods[lod] = {
-                vertexBuffer,
-                vertexCount: vertexData.vertexCount,
+                vertexBufferOpaque,
+                vertexCountOpaque: vertexDataOpaque.vertexCount,
+                vertexBufferTransparent,
+                vertexCountTransparent: vertexDataTransparent.vertexCount,
```
magcius: Technically you don't need two different buffers here; you can pack them into the same buffer and then use different start counts for each.

magcius (commented):
Please make sure you check one of the two boxes (I can take a guess given your past PRs, but I'd rather have confirmation).

src/RatchetAndClank/render-tfrag.ts
simon-paris: Fixed

src/RatchetAndClank/render-tfrag.ts
simon-paris: > how the second transparent draw is doing anything when you have the compareMode set to Greater; the second transparent draw will have the same positions as before, thus all the depth tests will fail

It's discarding based on alpha so the depth isn't written for transparent pixels.

The first draw draws pixels with alpha >0.75 with depth write.
The second draws pixels with alpha <0.75 without depth write.
Using Greater instead of GreaterEqual prevents the second draw from writing the same pixels again and double-blending them.

> I also don't see any alpha blending

It inherits the blend state from the scene megastate.

PR #941, opened 2026-07-17
Add Kingdom Hearts 3D: Dream Drop Distance and Kingdom Hearts Birth by Sleep

egartley:
Added worlds from the original versions of _Kingdom Hearts 3D: Dream Drop Distance_ and _Kingdom Hearts Birth by Sleep_. Both are included here since they use almost the exact same file formats and data structures (and hence most of the same code).

Credit goes to https://github.com/OpenKH/OpenKh for a majority of the parsing. OOT3D also helped as a reference for getting the animations working.

The two games have near feature parity; however, BBS is lacking behind a bit due to some complexities it has which Dream Drop doesn't. Therefore, I've kept BBS in experimental, but placed Dream Drop in the 3DS category since it's essentially complete aside from a handful of small fixes. These and other detailed notes are included in both of the scene files.

There's lots of things that can be configured in the UI panels, like which object sets are visible (enemies are hidden by default), as well as fog and texture application.

The only notable external change I made was moving the 3DS texture decoding from the Ocarina 3D folder into the common folder under `CTR` (renamed from `CTR_H3D`). It makes more sense to me to have it there, but I'm happy to move it back if there's a reason to. I've tested all the other 3DS games and they work as expected. There was also a GfxFormat I added back to account for variable vertex weight sizes in BBS.

As for setting up the data, here are the steps for Dream Drop:

- Dump the cart and extract romfs to `data/KingdomHeartsDDD_raw`
  - Only the `.rbin` files are needed, everything else can be discarded
- Create the `data/KingdomHeartsDDD` folder
- Run the extraction script `/src/KingdomHeartsDDD/tools/extract_rbin.py` (from within its folder)
  - The script needs a library called "ndspy" to decompress some files
  - Alternatively, the OpenKH tool could be used. Just need to run to it manually on the same ones listed in the script

The steps for BBS are little more complicated:

- Extract the disc contents
- Copy the 5 different `.DAT` files from `PSP_GAME/USRDIR` over to `data/KingdomHeartsBBS_raw`
- Some of these will need to be decrypted (3 and 4, iirc). There are kirk engine tools available that can do this
- Run the OpenKH tool called `OpenKh.Command.Bbsa` to extract their contents
- Combine the extracted contents into `data/KingdomHeartsBBS`
  - This should result in top-level directories such as `arc`, `CHARA` and `EFFECT`
- Repeat this process with the Japanese version of the game to get the files for the Jungle Book world
  - Only add the new files; existing ones with the same names should not be overwritten

image

image

## Pull Request Checklist


magcius:
Please supply data files to me through Discord or email.

src/KingdomHeartsDDD/render.ts
```diff
@@ -0,0 +1,134 @@
...
+            { location: DreamDropShader.a_Color, bufferIndex: DreamDropShader.a_Color, format: GfxFormat.F32_RGBA, bufferByteOffset: 0 },
+            { location: DreamDropShader.a_UV, bufferIndex: DreamDropShader.a_UV, format: GfxFormat.F32_RG, bufferByteOffset: 0 },
+        ];
+        const inVertexBufferDescriptors = [
+            { byteStride: 12, frequency: GfxVertexBufferFrequency.PerVertex },
+            { byteStride: 16, frequency: GfxVertexBufferFrequency.PerVertex },
+            { byteStride: 8, frequency: GfxVertexBufferFrequency.PerVertex }
+        ];
+        this.vertexBufferDescriptors = [
+            { buffer: createBufferFromData(cache.device, GfxBufferUsage.Vertex, GfxBufferFrequencyHint.Static, shape.vertices.map(v => v * scale).buffer), byteOffset: 0 },
```
magcius, line 109: I believe you're leaking a bunch of GPU resources? Have you used the leak checker? Set `main.loadSceneDelta = 0` in an F12 console, then press F9 to reload the scene and see what leaked.

egartley: Not seeing anything in the console after trying a few rooms at random. I had the device leak check function included at the top of the scene creation while working and fixed any as they came up. I just tested again with that as well and not seeing any.

image

Those buffers are destroyed here, in the inherited shape renderer class https://github.com/egartley/noclip.website/blob/45252ce5536c4b2f887680de778f3a645b06e0f1/src/KingdomHeartsDDD/lux.ts#L397

This is the top level destroy function, which will cascade down to the shapes (room -> model -> shape) https://github.com/egartley/noclip.website/blob/45252ce5536c4b2f887680de778f3a645b06e0f1/src/KingdomHeartsDDD/lux.ts#L816

PR #945, opened 2026-07-24
Donkey Kong 64: Implement Most Map Rendering Features

rmmh:
Overall, this brings the DK64 rendering not to 100% parity with the game rendering, but perhaps 90%. This was heavily based on the disassembly and partial decompilation in the [dk64 decompilation](https://gitlab.com/dk64_decomp/dk64).

- split DK64 archive into per-map, sharded resources, and common textures
- dynamic surfaces (water, lava, sand, etc)
- particle effects
- props and actors (trees, flowers, barrels)
- dynamic lighting 
- backdrops (gradient-based and textured)
- culling (chunk and object)

One DK64-unrelated change is optimizing the AABB check to use WASM SIMD128, based on it being a notable hit in the browser for all the objects I was rendering.

I recommend reviewing this file-by-file instead of commit-by-commit.

## Pull Request Checklist
- [X] Generative AI was used in the creation of this PR (for any part, including but not limited to coding, reverse engineering, or debugging). I have read [AI Contributions Policy](https://github.com/magcius/noclip.website/blob/main/README.md#ai-contributions-policy). All comments and documentation are human-authored.


rmmh:
demo: https://mod.ifies.com/f/260724_noclip_dk64/index.html#dk64/07;ShareData=A2JjZWl[?+Wk-B1WPBK)hI

src/DonkeyKong64/actor.ts
```diff
@@ -0,0 +1,443 @@
...
+    return { model, animation: null, animationSpeed: 0, renderer: 'skeletal' };
+}
+
+const warnedAnimationFeatures = new Set<string>();
+
+function warnAnimationFeatureOnce(feature: string, animationID: number, message: string): void {
+    if (warnedAnimationFeatures.has(feature))
+        return;
+    warnedAnimationFeatures.add(feature);
+    console.warn(`[DK64 actor] animation 0x${animationID.toString(16)}: ${message}`);
```
magcius: Please don't use globals.

src/DonkeyKong64/actor.ts
```diff
@@ -0,0 +1,443 @@
...
+    // TODO: Refer to func_global_asm_80619C2C, which also handles 0x0020
+    if ((flags & ~0x0002) !== 0) {
+        warnAnimationFeatureOnce(
+            'animation-flags',
+            animationID,
+            `unsupported header flags 0x${(flags & ~0x0002).toString(16)} are set`,
+        );
+    }
+    if (animationBoneCount > boneCount) {
+        warnAnimationFeatureOnce(
```
magcius: Does this fire at all?

src/DonkeyKong64/actor.ts
```diff
@@ -0,0 +1,443 @@
...
+    const toS16 = (value: number): number => (value << 16) >> 16;
+
+    for (let frame = 0; frame < frameCount; frame++) {
+        const frameRotations = new Int16Array(boneCount * 3);
+        let bitOffset = (frameDataStart + frame * frameStride) * 8 + initialBitOffset;
+        for (let bone = 0; bone < decodedBoneCount; bone++) {
+            for (let axis = 0; axis < 3; axis++) {
+                const descriptor = view.getUint16(0x14 + bone * 6 + axis * 2, false);
+                const bitCount = descriptor & 0x0F;
+                if (axis < 2 && descriptor !== 0) {
```
magcius: It shouldn't be hard to apply them, does this not fire?

src/DonkeyKong64/actor.ts
```diff
@@ -0,0 +1,443 @@
...
+    skeleton: ActorSkeleton,
+    sourceAnimation: ActorAnimation,
+    speed: number,
+    tick: number,
+    boneIndex: number,
+): void {
+    assert(boneIndex >= 0 && boneIndex < skeleton.offsets.length);
+    const pose = sampleActorAnimationPose(sourceAnimation, speed, tick);
+    buildActorBoneMatrices(skeleton, pose);
+    vec3.transformMat4(dst, boneOrigin, currentBoneMatrices[boneIndex]);
```
magcius: uhhh, so this samples all of the bones, then builds all of the bone matrices, only to pluck one out? Why not compute all of them in one go?

src/DonkeyKong64/actor.ts
```diff
@@ -0,0 +1,443 @@
...
+            boneParents: skeleton.parents,
+            sourceAnimation,
+            boundingBox: computeSkeletalAnimationBoundingBox(
+                sourcePositions, matrixIndices, skeleton.offsets, skeleton.parents,
+            ),
+        },
+    };
+}
+
+const currentBoneMatrices: mat4[] = [];
```
magcius: why are these globals

src/DonkeyKong64/actor.ts
```diff
@@ -0,0 +1,443 @@
...
+): SkeletalActorMesh | null {
+    const view = geometry.createDataView();
+    const runtimeBase = view.getUint32(0x00, false);
+    const displayListCount = view.getUint8(0x21);
+    const displayListTable = view.getUint32(0x04, false) - runtimeBase + 0x28;
+    const segmentBuffers: ArrayBufferSlice[] = [];
+    segmentBuffers[0x03] = geometry.slice(0x28);
+    if (displayListCount > 0) {
+        const firstDisplayList = view.getUint32(displayListTable, false) - runtimeBase + 0x28;
+        installActorVisibilitySegments(geometry, view, firstDisplayList, displayListTable, segmentBuffers);
```
magcius: visibility segments?

src/DonkeyKong64/actor.ts
```diff
@@ -0,0 +1,443 @@
...
+    animationID: number | null,
+    animationSpeed: number,
+    actor: SetupActor,
+    textureBuffers: ArrayBufferSlice[],
+    sharedOutput: RSPSharedOutput,
+): SkeletalActorMesh | null {
+    const view = geometry.createDataView();
+    const runtimeBase = view.getUint32(0x00, false);
+    const displayListCount = view.getUint8(0x21);
+    const displayListTable = view.getUint32(0x04, false) - runtimeBase + 0x28;
```
magcius: If this is an offset please rename to `displayListTableOffs`

src/DonkeyKong64/actor.ts
```diff
@@ -0,0 +1,443 @@
...
+        const pointer = view.getUint32(displayListTable + i * 4, false);
+        runDL_F3DEX2(state, 0x03000000 | (pointer - runtimeBase));
+    }
+    const output = state.finish();
+    if (output === null) {
+        console.warn(`[DK64 actor] skeletal model produced no draw calls`);
+        return null;
+    }
+    for (const drawCall of output.drawCalls)
+        drawCall.useVertexColors = false;
```
magcius: Do the models themselves not turn on G_SHADE? Is there a preamble that gets run normally that does that?

src/DonkeyKong64/actor.ts
```diff
@@ -0,0 +1,443 @@
...
+    tick: number,
+): void {
+    const pose = sampleActorAnimationPose(animation.sourceAnimation, animation.speed, tick);
+    buildActorBoneMatrices({
+        offsets: animation.boneOffsets,
+        parents: animation.boneParents,
+    }, pose);
+
+    for (let i = 0; i < animation.vertexCount; i++) {
+        const bone = Math.min(animation.matrixIndices[i], currentBoneMatrices.length - 1);
```
magcius: This is single-bone skinning, any reason not to do it on the GPU like we do in Banjo-Kazooie? I wouldn't mind CPU skinning but this doesn't feel intentional, per se.

src/DonkeyKong64/background.ts
```diff
@@ -0,0 +1,378 @@
...
+];
+
+const identityMatrix = mat4.create();
+const textureMatrix = mat4.create();
+
+function wrap01(v: number): number {
+    return v - Math.floor(v);
+}
+
+function calcAztecBeetleRaceTint(height: number): readonly [number, number, number] {
```
magcius: we have a Color class to use

src/DonkeyKong64/cull.ts
```diff
@@ -0,0 +1,330 @@
...
+
+interface MatrixAnimationCullData {
+    tracks: readonly MatrixAnimationCullTrack[];
+    sourcePositions: Float32Array;
+    vertexMatrixChains: readonly (readonly number[])[];
+    nodesByMatrixIndex: ReadonlyMap<number, MatrixAnimationCullNode>;
+    initialMatrices: ReadonlyMap<number, ReadonlyMat4>;
+}
+
+function computeConservativeMatrixAnimationBoundingBox(animation: MatrixAnimationCullData): AABB {
```
magcius: Huh? Why not compute the bounding box per-frame once you have the skeleton? Compute a bbox for each bone, and then once you have each bone posed you can compute your bbox for the object.

I feel like this is a solution that only an LLM would come up with :/

src/DonkeyKong64/cull.ts
```diff
@@ -0,0 +1,330 @@
...
+        } else {
+            cullGroup.boundingBox.union(cullGroup.boundingBox, boundingBox);
+        }
+        return cullGroup;
+    }
+
+    public setObjectCullBoundingBox(renderer: CullableRenderer, objectBoundingBox: AABB | null): void {
+        if (objectBoundingBox === null)
+            return;
+        renderer.setCullBoundingBox(objectBoundingBox);
```
magcius: Do you really have enough objects to need the cull groups? Typically that's only necessary if you're approaching 2000-3000 objects (e.g. the WoW renderer), otherwise you can just check the bbox against the frustum when rendering the object and early out there.

src/DonkeyKong64/f3dex2.ts
```diff
@@ -21,10 +21,45 @@ export enum RSP_Geometry {
...
+export interface AnimatedTexture {
+    segment: number;
+    group: number;
+    frameDuration: number;
+    frameOffset?: number;
+    crossfade?: boolean;
+    frames: ArrayBufferSlice[];
+}
+
+const animatedTextureSourceIDs = new WeakMap<ArrayBufferSlice, number>();
```
magcius: No globals.

src/DonkeyKong64/f3dex2.ts
```diff
@@ -66,11 +106,20 @@ export class RSPState {
     private DP_PrimColor = vec4.create();
     private DP_EnvColor = vec4.create();
     private DP_PrimLOD = 0;
+    private textureScrollSpeeds: number[] = [];
 
     public SP_MatrixIndex = 0;
+    private SP_MatrixChain: number[] = [];
+    private SP_MatrixStack: number[][] = [];
```
magcius: The heck am I looking at with both the matrix chain and the matrix stack?

src/DonkeyKong64/f3dex2.ts
```diff
@@ -148,16 +248,40 @@ export class RSPState {
...
+                },
+            };
         }
     }
 
-    private _flushTextures(dc: F3DEX.DrawCall): void {
-        // If textures are not on, then we have no textures.
-        if (!this.SP_TextureState.on)
+    private _flushTextures(dc: DrawCall): void {
+        // Actor 0x64 relies on G_TEXTURE staying active before its TMEM upload.
```
magcius: This comment feels a bit off? The condition you're adding a check for is that the DP_Combine don't use textures. We already check G_TEXTURE with the SP_TextureState.on.

src/DonkeyKong64/actor.ts
```diff
@@ -0,0 +1,443 @@
...
+    const animationFrame = ((tick * speed * animation.playbackRate) % animation.frameCount + animation.frameCount) % animation.frameCount;
+    const frame = Math.floor(animationFrame);
+    const nextFrame = (frame + 1) % animation.frameCount;
+    const t = animationFrame - frame;
+    const boneCount = animation.rotations[frame].length / 3;
+    const boneAngles: number[] = [];
+    for (let bone = 0; bone < boneCount; bone++) {
+        const a = animation.rotations[frame][bone * 3 + 2];
+        const b = animation.rotations[nextFrame][bone * 3 + 2];
+        boneAngles.push((a + (b - a) * t) * Math.PI * 2 / 0x10000);
```
magcius: `lerp(a, b, t) * MathConstants.TAU / 0x10000`

src/DonkeyKong64/light.ts
```diff
@@ -0,0 +1,471 @@
...
+            // from func_global_asm_8065C990
+            innerRadius: 300,
+            outerRadius: 1100,
+            color: [
+                light.color[0] * cameraFade,
+                light.color[1] * cameraFade,
+                light.color[2] * cameraFade,
+            ],
+            direction,
+            innerConeCos: Math.cos(light.innerAngle * Math.PI / 180),
```
magcius: MathConstants.DEG_TO_RAD

magcius (commented):
It's exhausting reading LLM-generated code. First pass.

rmmh:
Thanks for the first round of review. I apologize for the volume of code, I'll get rid of those global vars and try to condense it more before your next round. A few notes:

- hierarchical culling and computing static AABBs for animated objects was driven by some levels having thousands of props/actors and the cull testing dragging down framerate. I'll revisit this to see if it matters at all, and test it without the new SIMD aabb and drop that too if it's irrelevant.
- actor animation/skinning is incomplete. It's CPU-side because there are dynamic lights hanging on skinned lights, and I didn't want to redo a bunch of things to do dynamic lighting in vertex shaders.

magcius:
> It's CPU-side because there are dynamic lights hanging on skinned lights, and I didn't want to redo a bunch of things to do dynamic lighting in vertex shaders.

I assume this is a light attached to the end of a bone, or does it really hang off a skinned vertex? Computing the skeleton (aka bone matrices) on the CPU makes sense to me, just not the vertex positions.

src/DonkeyKong64/actor.ts
rmmh: No, deleted this and similar dead code in the actor file.

src/DonkeyKong64/actor.ts
rmmh: yes, but there aren't many lights that use this, and it's only done once per frame. computing just the single bone that the light is attached to requires duplicating a bunch of code.

src/DonkeyKong64/actor.ts
rmmh: fixed this and similar globals.

src/DonkeyKong64/actor.ts
rmmh: this does fire

src/DonkeyKong64/actor.ts
rmmh: done.

src/DonkeyKong64/background.ts
rmmh: done.

src/DonkeyKong64/actor.ts
rmmh: renamed to installDefaultActorPartSegments -- it's for replacing parts of models, like hands when they're holding different things

src/DonkeyKong64/cull.ts
rmmh: having a single AABB for the animated model lets me cull the entire thing without computing animation state, and that AABB is not much bigger than a more precise per-frame one

src/DonkeyKong64/cull.ts
rmmh: no, it's actually fine with release mode code-- dropped the hierarchical culling in favor of flat culling decisions.

src/DonkeyKong64/actor.ts
rmmh: but in our case it's only used for tag barrels-- in-game they only show the faces of chars you've unlocked

src/DonkeyKong64/actor.ts
magcius: Update all the bone matrices on the actor itself, then you'll have them to sample.

src/DonkeyKong64/cull.ts
magcius: Is computing the animation state that expensive? If the AABB is also not that different it would probably fine to keep the unanimated AABB for culling purposes.

src/DonkeyKong64/f3dex2.ts
rmmh: simplified in 870fb74c to be clearer

src/DonkeyKong64/f3dex2.ts
rmmh: clarified to "// If combiners are not reading textures, then we have no textures."

rmmh:
I responded to the first pass of comments, removed a bunch of dead code, removed the inspector tool, and generally tried to minimize the diff against main to improve reviewability and have a cleaner target to hopefully get squash-merged.

src/DonkeyKong64/cull.ts
rmmh: Simplified to a mostly-static AABB with some padding and a cheap translational check. Much simpler code, and can fix culling later if problems arise.

src/DonkeyKong64/actor.ts
rmmh: done in 367cf69c

src/DonkeyKong64/actor.ts
rmmh: double checked-- the prelude in func_global_asm_8063A968 sets G_SHADE and nothing clears it. fixing this makes props have shading they were missing before:

(before/after/ares emulator)
image
image
image

fixed in dc558932

src/DonkeyKong64/actor.ts
rmmh: done in a80da3d9

src/DonkeyKong64/parse.ts
```diff
@@ -0,0 +1,405 @@
...
+
+    private parseAnimatedTextures(view: DataView, animTexData: ArrayBufferSlice[]): void {
+        const table = view.getUint32(0x48, false);
+        const count = view.getUint32(table, false);
+        for (let i = 0; i < count; i++) {
+            const offs = table + 4 + i * 0x7C;
+            const frames: ArrayBufferSlice[] = [];
+            for (let j = 0; j < view.getUint8(offs + 3); j++) {
+                const frame = animTexData[view.getUint32(offs + 0x0C + j * 4, false)];
+                if (frame !== undefined)
```
magcius, line 300: Does this happen?

src/DonkeyKong64/scenes.ts
```diff
@@ -153,67 +304,126 @@ class DrawCallInstance {
...
             calcTextureMatrixFromRSPState(m, this.drawCall.SP_TextureState.s, this.drawCall.SP_TextureState.t, entry.width, entry.height, entry.tile.shifts, entry.tile.shiftt);
+            const speed = this.drawCall.textureScrollSpeeds[textureEntryIndex] ?? 0;
+            if (speed !== 0) {
+                const ticks = Math.floor(time / (1000 / 30));
+                if (ticks > 0) {
+                    const cycle = (Math.floor(255 / speed) + 1) * speed;
+                    const tileOffset = 255 - (((ticks - 1) * speed) % cycle);
+                    m[13] -= (tileOffset / 4) / entry.height;
+                }
+            }
```
magcius: I probably should clean this up but the math here feels overly complicated.

src/DonkeyKong64/scenes.ts
```diff
@@ -233,13 +443,41 @@ export class RenderData {
     public indexBufferDescriptor: GfxIndexBufferDescriptor;
     public vertexBufferData: Float32Array;
     public indexBuffer: GfxBuffer;
+    public vertexStart: number;
+    public indexStart: number;
+
+    constructor(device: GfxDevice, cache: GfxRenderCache, mesh: Mesh, dynamic = false) {
+        const sharedOutput = mesh.sharedOutput;
+        const drawCalls = mesh.rspOutput?.drawCalls ?? [];
+        this.indexStart = drawCalls.reduce(
```
magcius: Does it validly happen that you have indexStart != 0 and indexEnd != indices.length? Similarly, do you ever have vertexStart != 0 and vertexEnd != vertices.length? Feels like that's a bug in our F3DEX2 code if so... we shouldn't ever be outputting triangles we don't use.

src/DonkeyKong64/scenes.ts
```diff
@@ -268,30 +506,211 @@ export class RenderData {
...
+            const vertex = (this.firstVertex + i - vertexBufferFirstVertex) * 10;
+            if (!active) {
+                vertexBufferData[vertex + 0] = this.origin[0];
+                vertexBufferData[vertex + 1] = this.origin[1];
+                vertexBufferData[vertex + 2] = this.origin[2];
+                continue;
+            }
+            const rightOffset = this.halfWidth * SpriteBillboard.signs[i * 2];
+            const upOffset = this.halfHeight * SpriteBillboard.signs[i * 2 + 1];
+            vertexBufferData[vertex + 0] = centerX
```
magcius: Bizarre way to do this. Why not do this with the bone matrix? `calcBillboardMatrix`

src/DonkeyKong64/background.ts
```diff
@@ -0,0 +1,375 @@
...
+        renderInst.setMegaStateFlags({ depthCompare: GfxCompareMode.Always, depthWrite: false });
+
+        let offs = renderInst.allocateUniformBuffer(F3DEX_Program.ub_SceneParams, 16);
+        let mapped = renderInst.mapUniformBufferF32(F3DEX_Program.ub_SceneParams);
+        fillMatrix4x4(mapped, offs, identityMatrix);
+
+        const camera = viewerInput.camera.worldMatrix;
+        const pitch = Math.atan2(-camera[9], Math.hypot(camera[8], camera[10]));
+        const drawMatrix = mat4.create();
+        drawMatrix[13] = -2.5 * Math.sin(pitch);
```
magcius: calculating the angle only to go back and sin() it to me is weird but sure whatever. I think you can dot the camera forward vector against the world up vector, and since world up is 0, 1, 0, then I think this is just `Math.abs(camera[9])` ?

src/DonkeyKong64/cull.ts
```diff
@@ -0,0 +1,143 @@
...
+function computeDisplayListBoundingBox(sharedOutput: RSPSharedOutput, output: RSPOutput): AABB | null {
+    const boundingBox = new AABB();
+    let hasVertices = false;
+    for (const drawCall of output.drawCalls) {
+        const indexEnd = drawCall.firstIndex + drawCall.indexCount;
+        for (let index = drawCall.firstIndex; index < indexEnd; index++) {
+            const vertex = sharedOutput.vertices[sharedOutput.indices[index]];
+            boundingBox.min[0] = Math.min(boundingBox.min[0], vertex.x);
+            boundingBox.min[1] = Math.min(boundingBox.min[1], vertex.y);
+            boundingBox.min[2] = Math.min(boundingBox.min[2], vertex.z);
```
magcius: I really regret not making the N64 vertex have a vec3 position. If I do that then this will be `boundingBox.unionPoint(vertex.pos)`

src/DonkeyKong64/cull.ts
```diff
@@ -0,0 +1,143 @@
...
+            boundingBox.max[0] = Math.max(boundingBox.max[0], vertex.x);
+            boundingBox.max[1] = Math.max(boundingBox.max[1], vertex.y);
+            boundingBox.max[2] = Math.max(boundingBox.max[2], vertex.z);
+            hasVertices = true;
+        }
+    }
+    return hasVertices ? boundingBox : null;
+}
+
+export function computeBillboardBoundingBox(
```
magcius: Wouldn't a bounding sphere make more sense for this? You can already test containsSphere on a frustum.

src/DonkeyKong64/cull.ts
```diff
@@ -0,0 +1,143 @@
...
+        private sharedOutput: RSPSharedOutput,
+        private output: RSPOutput | null,
+        private animationBoundingBox: AABB | undefined,
+        private animationTranslationBounds: AABB | undefined,
+        private paddingRatio: number,
+    ) {
+    }
+
+    public getLocal(): AABB | null {
+        if (this.localBoundingBox === undefined) {
```
magcius: You should be able to compute this in the constructor, no? Also, why would you ever have a MeshBounds with null RSPOutput? That's just junk bounds, no?

src/DonkeyKong64/cull.ts
```diff
@@ -0,0 +1,143 @@
...
+            }
+            worldBoundingBox.transform(localBoundingBox, rootAnimation.baseMatrix);
+            worldBoundingBox.min[1] -= Math.abs(rootAnimation.positionYAmplitude);
+            worldBoundingBox.max[1] += Math.abs(rootAnimation.positionYAmplitude);
+        }
+        return worldBoundingBox;
+    }
+}
+
+export function computeSkeletalAnimationBoundingBox(
```
magcius: Again, I think you really want a bounding sphere here.

src/DonkeyKong64/f3dex2.ts
```diff
@@ -21,10 +21,45 @@ export enum RSP_Geometry {
...
+    private animatedTextureSourceIDs = new WeakMap<ArrayBufferSlice, number>();
+    private nextAnimatedTextureSourceID = 1;
+
+    public getAnimatedTextureCacheKey(frame: ArrayBufferSlice): number {
+        let sourceID = this.animatedTextureSourceIDs.get(frame);
+        if (sourceID === undefined) {
+            sourceID = this.nextAnimatedTextureSourceID++;
+            this.animatedTextureSourceIDs.set(frame, sourceID);
+        }
+        return 0x100000000 + sourceID;
```
magcius: Is this meaningful or are you just trying to find an unused space for the cacheKey parameter?

src/DonkeyKong64/f3dex2.ts
```diff
@@ -119,26 +165,81 @@ export class RSPState {
...
+            this.vertexCacheModelViewMatrixIndices[v0 + i] = this.matrixStack[this.matrixStack.length - 1] ?? [];
+            this.vertexCacheSourceAddresses[v0 + i] = dramAddr + i * 0x10;
         }
     }
 
-    private _translateTileTexture(tileIndex: number): number {
+    public gSPMatrix(dramAddr: number, matrixParams: number): void {
+        const segment = dramAddr >>> 24;
+        if (segment !== 0x04 && segment !== 0x09)
+            return;
```
magcius, line 220: Does this happen and should be ignored? Or should this be an assert. Wouldn't this upset the matrix stack size if we early return here?

src/DonkeyKong64/light.ts
```diff
@@ -0,0 +1,439 @@
...
+        });
+    }
+    for (const actor of setup.actors) {
+        const definition = getActorRenderDefinition(actor.type, 0);
+        if (definition === null || definition.lightBone === undefined || definition.animation === null)
+            continue;
+        const speed = definition.animationSpeed === 'setup' ? actor.lightSpeed : definition.animationSpeed;
+        lights.push({
+            kind: 'spot',
+            origin: vec3.fromValues((actor.position[0] + 0.3) * 3, actor.position[1] * 3, actor.position[2] * 3),
```
magcius: weird magic numbers?

src/DonkeyKong64/light.ts
```diff
@@ -0,0 +1,439 @@
...
+            // func_global_asm_8065EB10: vary light animation phase to avoid synchronization.
+            phase: prop.setupIndex,
+            maxDistance: maxDistance > 0 ? maxDistance : 700,
+        });
+    }
+    for (const actor of setup.actors) {
+        const definition = getActorRenderDefinition(actor.type, 0);
+        if (definition === null || definition.lightBone === undefined || definition.animation === null)
+            continue;
+        const speed = definition.animationSpeed === 'setup' ? actor.lightSpeed : definition.animationSpeed;
```
magcius: does lightSpeed differ from animationSpeed?

src/DonkeyKong64/light.ts
```diff
@@ -0,0 +1,439 @@
...
+    sectionVertexOffset: number,
+    chunk: RelitMapChunk | null,
+    lights: readonly DynamicLight[],
+): DynamicLighting | undefined {
+    if (chunk === null || !chunk.modulateVertexColors)
+        return undefined;
+
+    const animatedMaterialVertices = new Set<number>();
+    for (const drawCall of drawCalls) {
+        if (!drawCall.textureAnimations.some((animation) => animation !== undefined))
```
magcius: First off, why isn't this `drawCall.textureAnimations.length === 0`, why would the array have undefined values in it?

Second, wouldn't this catch any flipbook animations? Or are all of those assumed to be emissive.

src/DonkeyKong64/light.ts
```diff
@@ -0,0 +1,439 @@
...
+        camera[13] - light.origin[1],
+        camera[14] - light.origin[2],
+    ) / 3;
+    const distanceRatio = cameraDistance / light.maxDistance;
+    const cameraFade = distanceRatio < .8 ? 1 : Math.max(0, 1 - (distanceRatio - .8) / .2);
+    if (light.kind === 'spot') {
+        // func_global_asm_8069AB74: the light is hanging off the 3rd bone
+        light.pose.update(tick);
+        const boneMatrix = light.pose.boneMatrices[light.targetBone];
+        vec3.set(bonePosition, boneMatrix[12], boneMatrix[13], boneMatrix[14]);
```
magcius: getMatrixTranslation

src/DonkeyKong64/light.ts
```diff
@@ -0,0 +1,439 @@
...
+    let animationTick = (tick + light.phase) % totalDuration;
+    let keyframeIndex = 0;
+    while (animationTick >= keyframes[keyframeIndex].duration) {
+        animationTick -= keyframes[keyframeIndex].duration;
+        keyframeIndex++;
+    }
+    const current = keyframes[keyframeIndex];
+    const next = keyframes[(keyframeIndex + 1) % keyframes.length];
+    const t = animationTick / current.duration;
+    const radius = current.radius + (next.radius - current.radius) * t;
```
magcius: reinventing lerp from base principles

src/DonkeyKong64/light.ts
```diff
@@ -0,0 +1,439 @@
...
+        color: [
+            (current.color[0] + (next.color[0] - current.color[0]) * t) / 0xFF * intensity,
+            (current.color[1] + (next.color[1] - current.color[1]) * t) / 0xFF * intensity,
+            (current.color[2] + (next.color[2] - current.color[2]) * t) / 0xFF * intensity,
+        ],
+    };
+}
+
+function sampleLightAtPosition(light: ActiveLight, x: number, y: number, z: number): number {
+    const dx = x - light.origin[0];
```
magcius: this math would be a lot simpler phrased in vec3 form

src/DonkeyKong64/light.ts
```diff
@@ -0,0 +1,439 @@
...
+
+export function sampleObjectLighting(dst: vec3, lighting: ObjectLighting, activeLightCache: ActiveLightCache, enabled: boolean): vec3 {
+    if (!enabled)
+        return vec3.set(dst, 1, 1, 1);
+
+    vec3.copy(dst, lighting.ambientColor);
+    for (const dynamicLight of lighting.lights) {
+        const light = activeLightCache.get(dynamicLight)!;
+        const falloff = sampleLightAtPosition(light, lighting.origin[0], lighting.origin[1], lighting.origin[2]);
+        dst[0] += light.color[0] * falloff;
```
magcius: vec3.scaleAndAdd if you don't want to use Color for this.

src/DonkeyKong64/light.ts
```diff
@@ -0,0 +1,439 @@
...
+            direction,
+            innerConeCos: Math.cos(light.innerAngle * MathConstants.DEG_TO_RAD),
+            outerConeCos: Math.cos(light.outerAngle * MathConstants.DEG_TO_RAD),
+        };
+    }
+    const keyframes = light.animation;
+    const totalDuration = keyframes.reduce((sum, keyframe) => sum + keyframe.duration, 0);
+    let animationTick = (tick + light.phase) % totalDuration;
+    let keyframeIndex = 0;
+    while (animationTick >= keyframes[keyframeIndex].duration) {
```
magcius: kinda wish duration and things were precalculated and you had a helper for this (see the number of functions called `getAnimFrame` elsewhere)

src/DonkeyKong64/cull.ts
rmmh: fixed this to do this AABB.unionPoint properly

src/DonkeyKong64/cull.ts
rmmh: done

src/DonkeyKong64/cull.ts
rmmh: simplified to a static AABB

src/DonkeyKong64/f3dex2.ts
rmmh: meaningful to get unused space above 32-bit addresses, named the magic constant and commented

src/DonkeyKong64/light.ts
rmmh: fixed

src/DonkeyKong64/light.ts
rmmh: they're translated from the code-- 3x is world scale, 0.3 is an offset in the light setup. named and commented with sources

src/DonkeyKong64/background.ts
rmmh: nice, replaced with camera[9] (not abs, need offset to flip when looking down)

src/DonkeyKong64/light.ts
rmmh: done, totalDuration is precalculated now, and has getLightAnimFrame.

src/DonkeyKong64/scenes.ts
rmmh: Yes. Each mesh appends into one SharedOutput, so the indexStart/indexEnd selects that mesh's slice.

src/DonkeyKong64/light.ts
rmmh: cleaned up so that actors/attached lights can override animation speeds

src/DonkeyKong64/tools/extractor.ts
```diff
@@ -20,69 +30,849 @@ function determineSizeOfZlibStream(buffer: ArrayBufferSlice, srcOffs: number): n
...
+        const offs = (pointer & 0x7FFFFFFF) + PointerTableOffset;
+        const nextOffs = (nextPointer & 0x7FFFFFFF) + PointerTableOffset;
+        assert(nextOffs >= offs);
+        return this.romData.subarray(offs, nextOffs - offs);
+    }
+
+    public extractMapTable(): ArrayBufferSlice[] {
+        const MapData: ArrayBufferSlice[] = [];
+        for (let i = 0; i < MapCount; i++) {
             // TODO(jstpierre): Extract the proper size, and decompress on client.
```
magcius: This doesn't matter to me anymore, you can kill this TODO.

src/DonkeyKong64/tools/extractor.ts
```diff
@@ -20,69 +30,849 @@ function determineSizeOfZlibStream(buffer: ArrayBufferSlice, srcOffs: number): n
...
+}
+
+function globalAddressToOffset(globalASM: Buffer, address: number): number {
+    const offs = address - 0x805FB300;
+    assert(offs >= 0 && offs < globalASM.length);
+    return offs;
+}
+
+function parseSpriteData(globalASM: Buffer): SpriteInfo[] {
+    return Array.from({ length: 176 }, (_, i) => {
```
magcius: This is a very bizarre way to phrase this.

src/DonkeyKong64/tools/extractor.ts
```diff
@@ -20,69 +30,849 @@ function determineSizeOfZlibStream(buffer: ArrayBufferSlice, srcOffs: number): n
...
+//#endregion
+
+//#region resource sharding
+
+// Instead of having one big archive for all the DK64 content (~16MB),
+// split it into multiple archives.
+//
+// Structure:
+//        $MAP.crg1: map archive, unique mesh/texture/... data
+//   common_$N.crg1: a shard containing data used by multiple maps
```
magcius, line 600: I don't know if this common/sharding technique is worth the extractor complexity here. If you just pack the data used by each level, where does that get you on filesize?

src/DonkeyKong64/actors.ts
```diff
@@ -0,0 +1,308 @@
...
+
+const scratchVec3a = vec3.create();
+
+interface ActorAnimation {
+    playbackRate: number;
+    rotations: readonly Int16Array[];
+}
+
+interface ActorSkeleton {
+    offsets: vec3[];
```
magcius: ReadonlyVec3

src/DonkeyKong64/background.ts
```diff
@@ -0,0 +1,371 @@
...
+}
+
+export interface BackdropRenderer {
+    prepareToRender(renderInstManager: GfxRenderInstManager, viewerInput: Viewer.ViewerRenderInput): void;
+    destroy(device: GfxDevice): void;
+}
+
+const backdropWidth = 320;
+const backdropHeight = 240;
+const backdropVertexStride = 10; // pos.xyz, bone index, texcoord.st, color.rgba
```
magcius: Isn't this the normal RSP vertex stride?

src/DonkeyKong64/background.ts
```diff
@@ -0,0 +1,371 @@
...
+        const sourceRows = gradientBackdropSourceRows.get(paletteIndex) ?? defaultGradientSourceRows;
+        const sourceYToNDC = (sourceY: number): number => 1 - (sourceY - 336) / 120;
+        const vertices: number[] = [];
+        // the first/last colors extend past the gradient band offscreen
+        const rows = [
+            gradientOffscreenNDC,
+            ...sourceRows.map(sourceYToNDC),
+            -gradientOffscreenNDC,
+        ];
+        const colors = [palette[0], palette[0], palette[1], palette[2], palette[3], palette[3]];
```
magcius, line 334: So the NDC values here (for the default case are): 6, 3.8, -0.3, -0.8, -4.2, -6

NDC ranges from -1...1, so the top and bottom band are fully off-screen? Is this actually NDC, or is it in a different space?

src/DonkeyKong64/light.ts
```diff
@@ -0,0 +1,462 @@
...
+    const sinY = Math.sin(light.rotationY);
+    const cosY = Math.cos(light.rotationY);
+    const direction = vec3.fromValues(
+        cosY * bonePosition[0] + sinY * bonePosition[2] - spotLightOffsetX,
+        bonePosition[1],
+        -sinY * bonePosition[0] + cosY * bonePosition[2],
+    );
+    vec3.normalize(direction, direction);
+    return {
+        origin: light.origin,
```
magcius, line 359: Does the origin not get moved with the bone?

src/DonkeyKong64/light.ts
```diff
@@ -0,0 +1,462 @@
...
+const lightWorldScale = 3;
+// from func_global_asm_8065BAA0: point lights reach three times their keyframe radius.
+const pointLightOuterRadiusScale = 3;
+// from func_global_asm_8069AB74: createLight() uses actor X + 0.3 as the origin
+// and the animated position of bone 2 as the target.
+const spotLightOffsetX = 0.3;
+
+interface LightAnimationKeyframe {
+    intensity: number;
+    color: readonly [number, number, number];
```
magcius: Color? I would also accept ReadonlyVec3

src/DonkeyKong64/light.ts
```diff
@@ -0,0 +1,462 @@
...
+export function buildObjectLightingEnvironment(
+    vertexData: ArrayBufferSlice,
+    chunks: readonly RelitMapChunk[],
+    lights: readonly DynamicLight[],
+): ObjectLightingEnvironment {
+    const view = vertexData.createDataView();
+    return {
+        chunks: chunks.map((chunk) => {
+            const bounds = new AABB();
+            const end = Math.min(view.byteLength, chunk.vertOffset + chunk.vertSize);
```
magcius: Does this ever run off the end or is this untested safety

src/DonkeyKong64/props.ts
```diff
@@ -0,0 +1,1027 @@
...
+            const origin = vec3.fromValues(
+                prop.position[0] * worldScale,
+                prop.position[1] * worldScale,
+                prop.position[2] * worldScale,
+            );
+            mat4.translate(renderer.modelMatrix, renderer.modelMatrix, [
+                origin[0],
+                origin[1],
+                origin[2],
+            ]);
```
magcius:
```suggestion
const origin = vec3.scale(vec3.create(), prop.position, worldScale);
mat4.translate(renderer.modelMatrix, renderer.modelMatrix, origin);
```

(alternatively, `computeModelMatrixSRT` I think is what you want here)

src/DonkeyKong64/props.ts
```diff
@@ -0,0 +1,1027 @@
...
+    const terrainGrid = new TerrainTriangleGrid(terrainTriangles);
+    for (const prop of props) {
+        if (!propsByType.has(prop.type))
+            propsByType.set(prop.type, []);
+        propsByType.get(prop.type)!.push(prop);
+    }
+
+    for (const [propType, instances] of propsByType) {
+        // from func_global_asm_80636FFC: ignore these object types
+        if (propType === 0x0000 || propType === 0x0241)
```
magcius, line 991: Out of curiosity, do you know what this means?

src/DonkeyKong64/render.ts
```diff
@@ -0,0 +1,866 @@
...
+            if (mappings === undefined)
+                continue;
+            const textureAnimation = this.drawCall.textureBindings[i].animation!;
+            const frameDuration = Math.max(textureAnimation.frameDuration, 1);
+            const frameOffset = textureAnimation.frameOffset;
+            const frame = (Math.floor(animationFrame / frameDuration) + frameOffset) % mappings.length;
+            this.textureMappings[i] = mappings[frame];
+        }
+        if (sprites !== null) {
+            for (let i = 0; i < sprites.length; i++) {
```
magcius: I feel like DrawCallInstance might be losing the plot a little bit if it's being used for more than one draw call.

src/DonkeyKong64/props.ts
```diff
@@ -0,0 +1,1027 @@
...
+        public vertexOffsets: Uint32Array,
+        private sourcePositions: Float32Array,
+        private vertexModelViewMatrixIndices: number[][],
+        private initialMatrices: Map<number, mat4>,
+        public translationBounds: AABB,
+    ) {
+    }
+
+    private sample(tick: number): void {
+        for (const track of this.tracks) {
```
magcius, line 271: This whole function could use a scrubdown.

src/DonkeyKong64/render.ts
```diff
@@ -0,0 +1,866 @@
...
+            vec3.transformMat4(this.lookAtPosition, this.lookAtPosition, viewerInput.camera.viewMatrix);
+
+            mat4.lookAt(this.lookAtMatrix, Vec3Zero, this.lookAtPosition, Vec3UnitY);
+            offs += fillVec4(mappedF32, offs, this.lookAtMatrix[0], this.lookAtMatrix[4], this.lookAtMatrix[8]);
+            offs += fillVec4(mappedF32, offs, this.lookAtMatrix[1], this.lookAtMatrix[5], this.lookAtMatrix[9]);
+        }
+
+        const objectLightColor = this.objectLighting !== null
+            ? sampleObjectLighting(this.objectLightColor, this.objectLighting, activeLightCache, this.geometryData.dynamicLightingEnabled)
+            : null;
```
magcius: So this is weird. Either sampleObjectLighting returns 1, 1, 1 if it's not enabled, or we pass null here, which has the same effect. I don't know why we have two different ways to do the same thing.

src/DonkeyKong64/actors.ts
```diff
@@ -0,0 +1,308 @@
...
+
+export function getActorRenderDefinition(type: number, model: number): ActorRenderDefinition | null {
+    if (type === 0x10)
+        return { model: 0x81, animation: 0x402, animationSpeed: 'setup', lightBone: 2 };
+    if (type === 0x2A)
+        return { model: 0x97, animation: 0x402, animationSpeed: 'setup', lightBone: 2 };
+    if (model === 0)
+        return null;
+    // barrels yaw&bob, see func_global_asm_8068412C
+    if (type === 0x52 || type === 0x78 || type === 0x79)
```
magcius, line 119: I feel like I'd rather these get a subclass of the actor type which applies the yaw & bob in their update / prepareToRender method.

magcius (commented):
Looking better, still making my way through things.

src/DonkeyKong64/tools/extractor.ts
rmmh: ~4MB level loads instead of ~1MB loads.

src/DonkeyKong64/background.ts
rmmh: this is NDC, but fillDrawParams shifts the gradient entries based on the pitch angle

src/DonkeyKong64/light.ts
rmmh: nope! spotlights position are the actor origin and bone 2 is the target, oddly enough.

src/DonkeyKong64/props.ts
rmmh: 0x0000 is "test", 0x0241 is some stub zero-scale prop only found inside of torches-- maybe a marker for some lighting processing work (?)

src/DonkeyKong64/tools/extractor.ts
magcius: 4MB vs. 1MB is nothing. I'd rather do that and cut a bunch of the complexity here.

src/DonkeyKong64/f3dex2.ts
```diff
@@ -18,6 +18,9 @@ const G_MTX_PUSH = 0x04;
 // frames above 32-bit address space so they don't collide.
 export const animatedTextureCacheKeyBase = 0x100000000;
 
+// Standard RSP stride-- xyz, bone index, uv, rgba
```
magcius, line 21: do you really have to put the em dash in the comment? this should also be `VERTEX_STRIDE`

src/DonkeyKong64/f3dex2.ts
magcius: Still needs a response.

src/DonkeyKong64/props.ts
```diff
@@ -0,0 +1,1054 @@
...
+
+const transformComponentCount = 9;
+const scaleComponent = 0;
+const rotationComponent = 3;
+const translationComponent = 6;
+
+function interpolateTransform(node: PropAnimationNode, frame: number, t: number, component: number): number {
+    const current = node.transforms[frame * transformComponentCount + component];
+    const next = node.transforms[(frame + 1) * transformComponentCount + component];
+    return current + (next - current) * t;
```
magcius, line 174: Sorry, but three strikes and you're out.

PR #947, opened 2026-07-30
MPH: support animated textures & separate SP/MP map variants

rmmh:
Animated textures add a significant amount of movement to the scenes.

## Pull Request Checklist
- [X] Generative AI was used in the creation of this PR (for any part, including but not limited to coding, reverse engineering, or debugging). I have read [AI Contributions Policy](https://github.com/magcius/noclip.website/blob/main/README.md#ai-contributions-policy). All comments and documentation are human-authored.
- [ ] Generative AI was not used in the creation of this PR.


src/MetroidPrimeHunters/render.ts
```diff
@@ -130,7 +131,9 @@ class MaterialInstance {
     }
 
     public setOnRenderInst(template: GfxRenderInst, viewerInput: Viewer.ViewerRenderInput): void {
-        if (this.srt0Animator !== null) {
+        if (this.texCoordAnimator !== null) {
+            this.texCoordAnimator.calcTexMtx(scratchTexMatrix, this.material.texScaleS, this.material.texScaleT);
+        } else if (this.srt0Animator !== null) {
             this.srt0Animator.calcTexMtx(scratchTexMatrix, this.model.texMtxMode, this.material.texScaleS, this.material.texScaleT);
```
magcius: Can you remove the old srt0Animator code? It was copied from the G3D implementation and was never bound.

src/MetroidPrimeHunters/Scenes_MetroidPrimeHunters.ts
```diff
@@ -153,7 +154,7 @@ interface OBJI {
 }
 
 class MetroidPrimeHuntersSceneDesc implements Viewer.SceneDesc {
-    constructor(public id: string, public name: string, public arcName: string | null, public texName: string | null) {
+    constructor(public id: string, public name: string, public arcName: string | null, public texName: string | null, public sceneMode: MPHSceneMode = { kind: 'singlePlayer', area: 1 }, public modelId: string = id) {
```
magcius, line 157: Why is it area 1 by default? You never set any other area?

src/MetroidPrimeHunters/mph_anim.ts
```diff
@@ -0,0 +1,178 @@
...
+        const translationT = sampleChannelSmooth(view, this.animation.translationDataOffset, this.entry.translationT, frame, this.animation.frameCount, 4, true, translationWrap) / 0x1000;
+
+        mat2d.identity(dst);
+        mat2d.translate(dst, dst, [translationS * scaleS, translationT * scaleT]);
+        if (rotation !== 0) {
+            mat2d.translate(dst, dst, [0.5, 0.5]);
+            mat2d.rotate(dst, dst, rotation);
+            mat2d.translate(dst, dst, [-0.5, -0.5]);
+        }
+        mat2d.scale(dst, dst, [texScaleS * scaleS, texScaleT * scaleT]);
```
magcius: This matches the code in parseMaterial quite well, would be nice if it was shared.

src/MetroidPrimeHunters/mph_anim.ts
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    public calcTexMtx(dst: mat2d, texScaleS: number, texScaleT: number): void {
+        const view = this.animation.buffer.createDataView();
+        const frame = this.animationController.getTimeInFrames() % this.animation.frameCount;
+        const translationWrap = Math.floor(frame) === this.animation.frameCount - 1 ? 0x1000 : 0;
+        const scaleS = sampleChannelSmooth(view, this.animation.scaleDataOffset, this.entry.scaleS, frame, this.animation.frameCount, 4, true) / 0x1000;
+        const scaleT = sampleChannelSmooth(view, this.animation.scaleDataOffset, this.entry.scaleT, frame, this.animation.frameCount, 4, true) / 0x1000;
+        const rotation = sampleChannelSmooth(view, this.animation.rotationDataOffset, this.entry.rotation, frame, this.animation.frameCount, 2, false, 0x10000) / 0x10000 * Math.PI * 2;
+        const translationS = sampleChannelSmooth(view, this.animation.translationDataOffset, this.entry.translationS, frame, this.animation.frameCount, 4, true, translationWrap) / 0x1000;
```
magcius: I have a `fx32` helper for this which admittedly is probably dumb (it's just divide by / 0x1000), but it might be helpful for documentation.

src/MetroidPrimeHunters/mph_anim.ts
```diff
@@ -0,0 +1,178 @@
...
+}
+
+export class MPHTexCoordAnimator {
+    constructor(private animationController: AnimationController, private animation: MPHTexCoordAnimation, private entry: MPHTexCoordAnimationEntry) {
+    }
+
+    public calcTexMtx(dst: mat2d, texScaleS: number, texScaleT: number): void {
+        const view = this.animation.buffer.createDataView();
+        const frame = this.animationController.getTimeInFrames() % this.animation.frameCount;
+        const translationWrap = Math.floor(frame) === this.animation.frameCount - 1 ? 0x1000 : 0;
```
magcius: This is bizarre to me. Why do you only care about the translation wrap if we're on the last frame? We're not interpolating to frame 0 or anything.

src/MetroidPrimeHunters/mph_anim.ts
```diff
@@ -0,0 +1,178 @@
...
+        return getSample(channel.sampleIndex + sample);
+
+    const t = (subFrame << 12) >> shift;
+    const a = getSample(channel.sampleIndex + sample);
+    let b = getSample(channel.sampleIndex + nextSample);
+    if (!signed) {
+        if (a - b > 0x8000)
+            b += 0x10000;
+        else if (a - b < -0x8000)
+            return (((a + 0x10000) * (0x1000 - t) + b * t) >> 12) & 0xFFFF;
```
magcius: Please stop submitting PRs to me until you've looked through them and have at least guaranteed they don't contain "reinventing lerp from base principles".

src/MetroidPrimeHunters/mph_anim.ts
```diff
@@ -0,0 +1,178 @@
...
+    if (lastAlignedFrame <= frame)
+        return getSample(channel.sampleIndex + (frame - lastAlignedFrame) + (frame >> shift));
+
+    const sample = frame >> shift;
+    const nextSample = Math.min(sample + 1, channel.sampleCount - 1);
+    const subFrame = frame & (shift | 1);
+    if (subFrame === 0)
+        return getSample(channel.sampleIndex + sample);
+
+    const t = (subFrame << 12) >> shift;
```
magcius: Convert t to between 0...1, it will make the math much simpler here.

src/MetroidPrimeHunters/mph_anim.ts
```diff
@@ -0,0 +1,178 @@
...
+    const trackOffset = view.getUint32(texCoordTableOffset, true);
+    return {
+        texCoord: trackOffset !== 0 ? parseTexCoordAnimation(buffer, trackOffset) : null,
+    };
+}
+
+function sampleChannelFrame(view: DataView, dataOffset: number, channel: AnimationChannel, frame: number, frameCount: number, stride: number, signed: boolean): number {
+    const getSample = (index: number): number => {
+        const offset = dataOffset + index * stride;
+        if (stride === 4)
```
magcius, line 98: I guess stride of 4 is always signed? Can you at least assert on that?

src/MetroidPrimeHunters/mph_anim.ts
```diff
@@ -0,0 +1,178 @@
...
+
+    // From arm9.bin SampleFx32AnimationTrack @ 0x02058428 + SampleAngleAnimationTrack @ 0x02058344.
+    // Subsampled tracks interpolate until their aligned last frame, followed
+    // by full-rate tail samples.
+    const shift = channel.interpolation >>> 1;
+    const lastAlignedFrame = ((frameCount - 1) >> shift) << shift;
+    if (lastAlignedFrame <= frame)
+        return getSample(channel.sampleIndex + (frame - lastAlignedFrame) + (frame >> shift));
+
+    const sample = frame >> shift;
```
magcius: So what this actually is is the framerate in shift powers-of-two. So a shift of 1 means there's a sample every 2nd frame, and a shift of 2 means there's a sample every 4th frame.

And just so I understand this, if e.g. frameCount is 18, we have shift of 2 so we have a sample every 4 frames, that means we'll have samples for frames 0, 4, 8, 12, 16, but not for 20. Instead, we'll have a sample for 16, and then at the "tail" we'll have samples for 17 and 18. I think this could be easier to understand, both in the comments and in the code.

Also, please rename `sample` to `sampleIndex`

src/MetroidPrimeHunters/mph_anim.ts
```diff
@@ -0,0 +1,178 @@
...
+    const nextSample = Math.min(sample + 1, channel.sampleCount - 1);
+    const subFrame = frame & (shift | 1);
```
magcius: I think the right expression here is actually `frame & (shift - 1)`?

src/MetroidPrimeHunters/mph_anim.ts
```diff
@@ -0,0 +1,178 @@
...
+
+    const frame1 = (frame0 + 1) % frameCount;
+    const value1 = sampleChannelFrame(view, dataOffset, channel, frame1, frameCount, stride, signed);
+    if (wrap !== 0) {
+        let delta = value1 - value0;
+        delta -= Math.round(delta / wrap) * wrap;
+        return value0 + delta * t;
+    }
+
+    return value0 + (value1 - value0) * t;
```
magcius: make it stop bleeding

src/MetroidPrimeHunters/mph_binModel.ts
```diff
@@ -438,6 +446,9 @@ export function parseMPH_Model(buffer: ArrayBufferSlice): MPHbin {
             const name = node.name;
             const jointMatrix = node.transform;
             nodes.push({ name, jointMatrix });
+            // RenderEnabledModelNodes @ 0x02047808 uses meshID to index the
+            // interleaved material/shape ushort table.
+            nodeMeshRanges.push({ meshStart: node.meshID >> 1, meshCount: node.meshCount });
```
magcius: `>>>` for unsigned shifts

src/MetroidPrimeHunters/render.ts
magcius, line 153: This is not on you, and you don't need to fix this, but I really would like to stop using mat2d/mat3 and just use mat4 in our code going forward, using `fillMatrix4x2` to pack into the UBO.

src/MetroidPrimeHunters/mph_anim.ts
rmmh: We actually are in the smooth sampling. This avoids interpolating between the last and the first texture coordinates for a few frames. Redone to be cleaner.

src/MetroidPrimeHunters/mph_anim.ts
rmmh: There wasn't an fx32 lerp, and I wasn't sure that the result would be equivalent enough in floating point! But it looks fine in FP with lerps.

src/MetroidPrimeHunters/Scenes_MetroidPrimeHunters.ts
rmmh: singlePlayer is always area 1, yes. in particular, it means that it won't show nodes marked _s02 through _s08 in single player mode.

src/MetroidPrimeHunters/mph_anim.ts
```diff
@@ -0,0 +1,178 @@
...
+    if (t === 0)
+        return value0;
+
+    const frame1 = (frame0 + 1) % frameCount;
+    const value1 = sampleChannelFrame(view, dataOffset, channel, frame1, frameCount, stride, signed);
+    if (wrap !== 0) {
+        let delta = value1 - value0;
+        delta -= Math.round(delta / wrap) * wrap;
+        return lerp(value0, value0 + delta, t);
+    }
```
magcius: I think this is equivalent to `lerpAngle(value0, value1, t, wrap)`? lerpAngle is really a modulo wrap like you have here. It's probably more confusing to replace what you have with lerpAngle, so don't really care either way.

src/MetroidPrimeHunters/mph_anim.ts
```diff
@@ -0,0 +1,178 @@
...
+    // From arm9.bin SampleFx32AnimationTrack @ 0x02058428 + SampleAngleAnimationTrack @ 0x02058344.
+    // The interpolation value stores the sample rate as a power-of-two shift:
+    // shift 1 has samples every two frames, shift 2 every four frames, and so on.
+    // Regular subsampled tracks are followed by full-rate samples for the tail.
+    // For example, an 18-frame track with shift 2 has regular samples at frames
+    // 0, 4, 8, 12, and 16, followed by tail samples at frames 17 and 18.
+    const sampleShift = channel.interpolation >>> 1;
+    const lastAlignedFrame = ((frameCount - 1) >> sampleShift) << sampleShift;
+    if (lastAlignedFrame <= frame)
+        return getSample(channel.sampleIndex + (frame - lastAlignedFrame) + (frame >> sampleShift));
```
magcius:
```suggestion
    const frameSampleIdx = frame >> sampleShift;
    const lastKeyframe = ((frameCount - 1) >> sampleShift) << sampleShift;
    if (frame >= lastKeyframe) {
        assert(frameSampleIdx === (lastKeyframe >> sampleShift));
        return getSample(channel.sampleIndex + frameSampleIdx + (frame - lastKeyframe));
    }
```

src/MetroidPrimeHunters/mph_anim.ts
```diff
@@ -0,0 +1,178 @@
...
+
+    const sampleIndex = frame >> sampleShift;
+    const nextSampleIndex = Math.min(sampleIndex + 1, channel.sampleCount - 1);
```
magcius: This min shouldn't happen, because the keyframe tail above should take care of it, no? I would add an assert here for this.

src/MetroidPrimeHunters/mph_anim.ts
```diff
@@ -0,0 +1,178 @@
...
+
+export interface MPHAnimation {
+    texCoord: MPHTexCoordAnimation | null;
+}
+
+function parseChannel(view: DataView, entryOffset: number, interpolationOffset: number, countOffset: number, indexOffset: number): AnimationChannel {
+    return {
+        interpolation: view.getUint8(entryOffset + interpolationOffset),
+        sampleCount: view.getUint16(entryOffset + countOffset, true),
+        sampleIndex: view.getUint16(entryOffset + indexOffset, true),
```
magcius: I think renaming this to firstSampleIndex might help clarify some code below.

magcius (commented):
Just a couple more notes, this is looking great now. I think none of them have to be applied, if you like the code you have more, just reply and I'll merge as-is.

src/MetroidPrimeHunters/Scenes_MetroidPrimeHunters.ts
magcius: hmm, where are _s02 through _s08 used if not through single-player?

src/MetroidPrimeHunters/mph_anim.ts
magcius: I realize now this expression isn't it (I meant to write `frame & ((1 << shift) - 1)`), but I still don't quite understand the wisdom of `frame & (shift | 1)`

rmmh: No. This is translated from the game:

```
mov  r2, r2, asr #1   ; sampleShift
  orr  r3, r2, #1       ; mask = sampleShift | 1
  ands ..., frame, r3
```

It's either a typo or maaaybe a peephole optimization-- x|1 == (1<<x)-1 for x=1 and x=2, which are the only values the game has here.

Either way, changed to `const subFrame = frame & ((1 << sampleShift) - 1);` for clarity.

src/MetroidPrimeHunters/Scenes_MetroidPrimeHunters.ts
rmmh: It isn't always 1 in-game, and selects variants for assets that are reused in multiple areas, but that additional assignment is in a future commit with a richer areaMetadata table.

rmmh:
merge at will

src/MetroidPrimeHunters/mph_anim.ts
magcius: Why did you mark this resolved?

rmmh: https://github.com/magcius/noclip.website/pull/947/changes#diff-42683bfc1e5d891ed17f7486ccf0b366726fcd5acff5989ba11a0a611ffb6e2dR94

magcius: Thanks. Didn't spot it.

PR #966, opened 2026-09-12
Fix bmodels being rendered at the wrong offset

L-P:
Instead of dumping bmodels at the origin, offset them by their entity origin.  

I could not get shared bmodels to work but that's a "recent" hack only permitted by the *HLT family of compilers and not relevant to commercially released GoldSrc games.

Before/after:

## Pull Request Checklist


magcius:
yeah this works, though if we wanted entities to be dynamic in the future like in Source, then we'd need to have the bmodels be separate renderers that the entities manipulate. seems good for now though

PR #972, opened 2026-09-14
Implement GoldSrc rendermode

L-P:
I'm not too happy about the fragment shader but setting render modes via defines would require a substantial refactoring to duplicate bmodels per entity at parsing time.  
Proper per face sorting would also require to split these bmodels per face.

A worst case scenario is visible in [c1a0b](<https://noclip.website/#HalfLife/c1a0b;ShareData=AfQ2g+7(=6+i7s@+wuYegn>) where you have convex bmodels without an origin placed in a row.

2026-09-14_11-32 png

## Pull Request Checklist


src/Common/IdTech2/Render.ts
```diff
@@ -537,6 +596,56 @@ class LightmapManager {
...
+enum RenderMode {
+    Normal = 0,
+    Color,
+    Texture,
+    Glow, // only valid on sprites
+    Solid,
+    Additive,
+}
+
+class RenderOptions {
```
magcius, line 608: don't love the way this is implemented, not sure why RenderOptions is a class. it also doesn't use color at all in the class?

src/Common/IdTech2/Render.ts
```diff
@@ -537,6 +596,56 @@ class LightmapManager {
...
+    blendMode(): Partial<AttachmentStateSimple> {
+        if (this.mode === RenderMode.Additive) {
+            return {
+                blendMode: GfxBlendMode.Add,
+                blendSrcFactor: GfxBlendFactor.SrcAlpha,
+                blendDstFactor: GfxBlendFactor.One,
+            };
+        }
+
+        return {
```
magcius, line 637: solid should be Add/One/Zero

also you can combine transparent/additive if you use Add/One/OneMinusSrcAlpha, multiply output .a into .rgb in the shader, and set .a to 0 for additive case

src/Common/IdTech2/Render.ts
```diff
@@ -603,6 +712,12 @@ export class BSPRenderer {
                     continue;
             }
 
+            if (this.context.isGoldSrc) {
+                // In Half-Life some ladders are not using tool textures (eg. in c2a4).
```
magcius, line 716: the entity code itself probably hides the model here

src/Common/IdTech2/Render.ts
```diff
@@ -611,26 +726,43 @@ export class BSPRenderer {
...
+            };
+
+            const renderOptions = RenderOptions.fromBSPEntity(entity);
+            setAttachmentStateSimple(megaStateFlags, renderOptions.blendMode());
+            template.setMegaStateFlags(megaStateFlags);
+
+            let offs = template.allocateUniformBuffer(GoldSrcProgram.ub_SceneParams, 16+4);
+            const d = template.mapUniformBufferF32(GoldSrcProgram.ub_SceneParams);
+            offs += fillMatrix4x4(d, offs, view.clipFromWorldMatrix);
+            offs += fillVec3v(d, offs, view.eyePos, view.time);
```
magcius, line 751: don't create a scene params per model, if you want another template to put the render state on if you don't want the model renderer to handle it for some reason, you can just push another template

src/Common/IdTech2/Render.ts
```diff
@@ -701,6 +833,14 @@ export class IdTech2Renderer implements SceneGfx {
                 this.mainView.mainList.drawOnPassRenderer(this.renderHelper.renderCache, passRenderer);
             });
         });
+        builder.pushPass((pass) => {
```
magcius, line 836: this doesn't need to be in a separate pass, you can render it in Main after the main list

magcius (changes requested):

src/Common/IdTech2/Render.ts
L-P: It was an interface and scattered functions initially but it felt cleaner to put everything in a class.  
I come from a language where there's no distinction between the two. I can split it up again if you want.

The color is only used in the shader.

src/Common/IdTech2/Render.ts
L-P: Yes `func_ladder` are forcibly set to rendermode texture and renderamt 0 when they're precached in the engine. Which is dubious as most of the HL game logic code is.

src/Common/IdTech2/Render.ts
L-P: Removed.

src/Common/IdTech2/Render.ts
L-P: Done. I left the shader as-is, seems simpler that way.

src/Common/IdTech2/Render.ts
L-P: Done.

src/Common/IdTech2/Render.ts
```diff
@@ -409,19 +460,37 @@ class BSPModelRenderer {
...
         if (!this.visible)
             return;
 
+        const template = renderInstManager.pushTemplate();
+        const renderOptions = RenderOptions.fromBSPEntity(entity);
+        const megaStateFlags: Partial<GfxMegaStateDescriptor> = {};
+        setAttachmentStateSimple(megaStateFlags, renderOptions.blendMode());
+        template.setMegaStateFlags(megaStateFlags);
+
+        if (renderOptions.isTranslucent()) {
```
magcius, line 474: much better. since you already pass the renderOptions into the model renderer you could do all this stuff in here but it's fine

magcius (approved):

PR #975, opened 2026-09-15
Add render options to IdTech2

L-P:
1. Disable texture filtering by default in IdTech2 This should not be too controversial for Quake as texture filtering only appeared with GLQuake one year after the initial release.  
   I kept in on by default on Half-life as it's probably the way most people first experienced the game even though it's clearly an afterthought when you look at alphatest textures.

2. The trigger display toggle is a freebie. It helped me debug things so I kept it in.

Tested on both Quake and Half-Life.

2026-09-15_19-49 png

## Pull Request Checklist


src/Common/IdTech2/Render.ts
```diff
@@ -847,4 +887,38 @@ export class IdTech2Renderer implements SceneGfx {
...
+
+    private createRenderSettingsPanel(): UI.Panel {
+        const panel = new UI.Panel();
+        panel.customHeaderBackgroundColor = UI.COOL_BLUE_COLOR;
+        panel.setTitle(UI.RENDER_HACKS_ICON, 'Render Settings');
+
+        const filteringCheckbox = new UI.Checkbox('Enable texture filtering', this.context.enableTextureFiltering);
+        filteringCheckbox.onchanged = () => {
+            this.context.enableTextureFiltering = filteringCheckbox.checked;
+            this.rebuild();
```
magcius, line 911: blindly rebuilding is going to leak things. why does this need to be a full rebuild instead of skipping over entities / model renderers at runtime?

magcius (changes requested):

src/Common/IdTech2/Render.ts
L-P: I was kinda wary about that, it was for the sake of making a smaller change as
currently the filtering is done only once before rendering.  

I'll try to move things around and filter things at runtime.

